PR #35, opened 2024-01-24
[Bugfix] assertions in debug mode fixed

### LINKED ISSUE #34: Cannot compile if CMAKE_BUILD_TYPE is not set or set to Debug
When building with `CMAKE_BUILD_TYPE` unset or set to `Debug`, I got the following error:

```
/local/home/user/Projets/kokkos-fft/fft/src/KokkosFFT_Helpers.hpp(27): error: no instance of function template "KokkosFFT::Impl::is_out_of_range_value_included" matches the argument list
            argument types are: (std::vector<int, std::allocator<int>>, const int)
   !KokkosFFT::Impl::is_out_of_range_value_included(axes, rank)
```

yasahi-hpc:
Fixes #34


yasahi-hpc:
@pzehner Now is OK for you?

pzehner:
Sounds good to me.

PR #29, opened 2024-01-23
Add CI for GPU tests on Ruche

pzehner:
This PR aims to add execution of GPU tests on the Ruche supercomputer.

PR #40, opened 2024-01-30
Fix CMake scripts for installation and SYCL backend

pzehner:
This PR aims to fix two aspects of the CMake setup:

1. Installation of the library with `cmake --install`/`make install`: this PR uses [standard CMake practices](https://cmake.org/cmake/help/git-stage/manual/cmake-packages.7.html#config-file-packages) to make the project configured with `CMAKE_INSTALL_PREFIX=/path/to/install` findable by a project configured with `CMAKE_PREFIX_PATH=/path/to/install`; and
2. Installation of the library with a SYCL backend: this PR adds missing CMake commands for this backend.


yasahi-hpc:
@pzehner Maybe we also need to add `install_test/bin/install_sycl.sh` and add install test in workflow. But these stuffs will eventually be integrated differently in your new CI, so it may not make much sense.

cedricchevalier19 (approved):
So kokkosFFT is the official name ?

I have a few comments about the build scripts, but not from this pr: you should not pass Kokkos Cmake options except for building Kokkos. When using an external and already installed kokkos, you should check the configuration from the actual install.

yasahi-hpc:
As for the name, I propose kokkos-fft for docs (including repo name) and KokkosFFT for the rest (source codes, CMake). This may be the common way in Kokkos related libraries. 

The build procedure would be updated in new [PR for CI](#29) by @pzehner ?

pzehner:
@yasahi-hpc yes, I didn't add an `install_test/bin/install_sycl.sh` script as this work is handled differently in #29.

About the naming convention, at least on CMake side, I followed the same approach of Kokkos, for which the repository name is `kokkos` (lower case), and the CMake name is `Kokkos` (Pascal case, or at least title case).

yasahi-hpc:
@pzehner Sure. If you are fine, I will merge this.

pzehner:
I'm OK for merging.

PR #42, opened 2024-02-05
Add doxygen

yasahi-hpc:
In this PR, I have added a template for documentation with doxygen + sphinx. Workflow may be added based on this PR.


docs/Makefile
jbigot, line 1: Why is a Makefile needed in addition to cmake?

docs/conf.py
```diff
@@ -0,0 +1,99 @@
...
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+import subprocess, os
+
+def configureDoxyfile(input_dir, output_dir, doxyfile_in, doxyfile_out):
```
jbigot, line 18: this same functionality is present in CMake, why is there a duplication?

docs/conf.py
```diff
@@ -0,0 +1,99 @@
...
+    ###                        -DKokkosFFT_ENABLE_DOCS=ON ..'
+    ###build = 'cmake --build . -j 2'
+    ###                        
+    ###subprocess.call(f'cd ../; mkdir build; cd build; {cmake_commands}; {build}', shell=True)
+    input_dir = cwd + '/..'
+    output_dir = cwd +'/doxygen/'
+    doxyfile_in = cwd + '/Doxyfile.in'
+    doxyfile_out = cwd + '/Doxyfile'
+    configureDoxyfile(input_dir, output_dir, doxyfile_in, doxyfile_out)
+    subprocess.call('pwd; ls -lat; doxygen Doxyfile; ls -lat doxygen/xml', shell=True)
```
jbigot, line 63: here again I don't understand, does CMake handle Doxygen call or python?

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -39,34 +45,72 @@ namespace Impl {
...
+
+  //! directions of fft
   KokkosFFT::Impl::Direction m_direction;
 
-  // Keep extents
+  //! extents of input/output views
   extents_type m_in_extents, m_out_extents;
 
-  // Only used when transpose needed
+  //! Internal buffers used for transpose
```
jbigot, line 101: without `@{` `@}`, you only comment the first next variable `m_in_T`, not the following one `m_out_T`

jbigot (approved):
Great documentation effort! This looks great to me.

docs/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+set(DOXYFILE_OUT ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile)
+
+# Set the Doxygen input and output directories in the Doxyfile
+configure_file(${DOXYFILE_IN} ${DOXYFILE_OUT})
+
+# Doxygen won't create this for us
+file(MAKE_DIRECTORY ${DOXYGEN_OUTPUT_DIR}) 
+
+# Only regenerate Doxygen when the Doxyfile or public headers change
+add_custom_command(OUTPUT ${DOXYGEN_INDEX_FILE}
```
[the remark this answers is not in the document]
yasahi-hpc: I did not notice this command. Seems simpler.

docs/Makefile
yasahi-hpc: It is for sphinx, and in the end, we do not rely on CMake in the rtd workflow.

docs/conf.py
yasahi-hpc: In the rtd workflow, only the python works. CMake is not used.

fft/src/KokkosFFT_Plans.hpp
yasahi-hpc: So, we need the comment like this?

`//@{ //! Internal buffers used for transpose @}//`

jbigot: not exactly, the `@{` and `@}` should surround the variables, see https://www.doxygen.nl/manual/grouping.html

yasahi-hpc: So, 

```
///@{ 
  //! extents of input/output views
  extents_type m_in_extents, m_out_extents;
///@}
```

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -39,34 +45,72 @@ namespace Impl {
...
   nonConstInViewType m_in_T;
   nonConstOutViewType m_out_T;
```
jbigot, line 103:
```suggestion
  //! @{
  //! Internal buffers used for transpose
  nonConstInViewType m_in_T;
  nonConstOutViewType m_out_T;
  //! @}
```

PR #44, opened 2024-02-08
Improve CI

pzehner:
This PR aims to improve the CI workflow by the following elements:

- [x] Rename images after their compiler environment, and explicitly use their name (rather than relying on backend name);
- [x] Increase tests verbosity in case of failure (as CTest only shows outcomes by default);
- [x] Rebuild base images on pushes on the `main` branch if Docker images were modified;
- [x] Use tags for base images: main images are always tagged `latest` and PR images are tagged with their commit hash; and
- [x] Update ROCM image (remove tag that was introduced by mistake).

Post-merge actions:

- [ ] Remove images with old names.


.github/workflows/build_test.yaml
```diff
@@ -206,18 +212,18 @@ jobs:
...
+          singularity run --nv --bind $PWD/build:/work/build -H /work/build base_${{ matrix.backend.image }}_singularity_${{ needs.check_docker_files.outputs.image_name_suffix }}_latest.sif \
+          ctest --verbose
         if: ${{ matrix.backend.name == 'cuda' }}
 
       - name: Run OpenMP tests within Docker image
         run: |
-          docker run -v $PWD/build:/work/build -w /work/build ghcr.io/cexa-project/kokkos-fft/base_${{ matrix.backend.name }}_${{ needs.check_docker_files.outputs.image_name_suffix }} \
-          ctest
+          docker run -v $PWD/build:/work/build -w /work/build ghcr.io/cexa-project/kokkos-fft/base_${{ matrix.backend.image }}_${{ needs.check_docker_files.outputs.image_name_suffix }} \
+          ctest --verbose
```
cedricchevalier19, line 232: You can also use

```suggestion
          ctest --output-on-failure
```

pzehner: Better indeed.

yasahi-hpc (approved):
I think this PR helps to introduce threads backend. Thanks a lot! 
The image names may be changed in the future.

PR #45, opened 2024-02-09
Fix missing points in CI

pzehner:
As an addendum to #44, this PR finalizes two improvements on the CI:

- [x] Fix old names of images for cleanup; and
- [x] Allow to generate and cleanup base images manually (through GitHub).


yasahi-hpc (approved):
Thank you for this PR.
I will merge this. 

By the way, do you have any idea to test the workflow of auto cleanup capability?

PR #46, opened 2024-02-09
Add threads backend

### LINKED ISSUE #31: Looks for FFTW when Kokkos::Threads is enabled
It would be nice to modify the config to include the use case of Kokkos::Threads, it works fine and is actually used commonly on MacOS since OpenMP is not available by default.

yasahi-hpc:
Resolves #31

In this PR, we add the `Kokkos::Threads` backend for broader users.
1. CMake for build and installation updated
2. Add CI for `Kokkos::Threads` and `Kokkos::Serial` backends


yasahi-hpc:
@pzehner it seems fine. Could you please review this PR?

pzehner:
Everything ran fine, you adapted the CI well! I'm just a bit surprised that the serial build benchmark execution time (152 s) is of the same order of magnitude as the OpenMP benchmark (147 s) and the CUDA benchmark (161 s).

pzehner (approved):

yasahi-hpc:
Thanks!

As for the benchmark, it is not that surprising. cufft plan creation is very time consuming, which is even larger than fft itself. This is why we need a reuse plan mode. I did not added these benchmarks yet. I will add more benchmarks and do performance tests next week.

PR #50, opened 2024-02-13
[WIP] Add CI with nvhpc

yasahi-hpc:
This PR aims at adding CI for nvhpc environment. 
1. Docker file for nvhpc environment added
2. CI for build and test on nvhpc environment added

### DIFF
--- a/docker/nvhpc/Dockerfile
+++ b/docker/nvhpc/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE=nvcr.io/nvidia/nvhpc:23.11-devel-cuda_multi-ubuntu22.04
+ARG BASE=nvcr.io/nvidia/nvhpc:24.1-devel-cuda12.3-ubuntu22.04
 FROM $BASE
 
 ARG ADDITIONAL_PACKAGES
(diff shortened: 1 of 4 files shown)


docker/nvhpc/Dockerfile
```diff
@@ -0,0 +1,47 @@
+ARG BASE=nvcr.io/nvidia/nvhpc:23.11-devel-cuda_multi-ubuntu22.04
```
pzehner, line 1: This image is 9.3 GB (see [NGC catalog](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/nvhpc/tags))! I think it would be better to use this image `nvcr.io/nvidia/nvhpc:23.11-devel-cuda12.3-ubuntu20.04` instead, which is 5.51 GB. It would give a similar environment as the NVCC image.

yasahi-hpc: As you expected, I ran out of disk space. Try again with a smaller image.

yasahi-hpc:
Close this since Kokkos does not support nvhpc as a CUDA compiler any more

PR #64, opened 2024-02-20
Install test

### LINKED ISSUE #43: Pre-build Kokkos on CI
Build and install Kokkos on Release mode in the images.

Kokkos would be discovered as an installed library, backend parameters should be deduced and not passed again.

The test example build at the end of the workflow should properly use this installed version of Kokkos.

Optionally, `KokkosFFT_INTERNAL_Kokkos` should be removed.

### LINKED ISSUE #66: Pertinence of Kokkos_Version_Check
I think the feature in `cmake/Kokkos_Version_Check.cmake` should be easily replaced by the CMake standard [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature) version option:

```cmake
find_package(Kokkos 4.2.0 REQUIRED)
```

Maybe we don't need to specify the version up to the patch level (i.e. `4.2` should be enough).

Otherwise, the `cmake/Kokkos_Version_Check.cmake` is invalid and should be:

```diff
diff --git a/cmake/Kokkos_Version_Check.cmake b/cmake/Kokkos_Version_Check.cmake
index 3f3f5cb..98a38f5 100644
--- a/cmake/Kokkos_Version_Check.cmake
+++ b/cmake/Kokkos_Version_Check.cmake
@@ -1,11 +1,11 @@
 function(check_minimum_required_kokkos kokkos_required_version)
-    if(${Kokkos_VERSION} STREQUAL "")
+    if(Kokkos_VERSION STREQUAL "")
         message(FATAL_ERROR "Kokkos_VERSION not set. Cannot check Kokkos satisfies the minimum required version.")
     else()
-        if(${Kokkos_VERSION} VERSION_GREATER_EQUAL ${kokkos_required_version})
+        if(Kokkos_VERSION VERSION_GREATER_EQUAL ${kokkos_required_version})
             message(STATUS "Found Kokkos version ${Kokkos_VERSION} at ${Kokkos_DIR}")
         else()
             message(FATAL_ERROR "Kokkos FFT ${KOKKOSFFT_VERSION} requires ${kokkos_required_version} or later.")
         endif()
     endif()
```

yasahi-hpc:
Resolves #43
Resolves #66 

This PR aims at recovering CI for using KokkosFFT as subdirectory or library.

1. Build app using KokkosFFT as a subdirectory in CMake project
2. Build app using Kokkos and KokkosFFT as libraries in CMake project
3. Add these tests in CI
4. fix CMake issues when using KokkosFFT as a subdirectory
5. docs would be updated based on the changes


pzehner (changes requested):
Based on our previous discussion, testing the installation of KokkosFFT should be done in a different job:

```yaml
install:
  # same `needs` and `if` as the `build` job.
  # the `strategy` may be similar, without testing host-device configuration (i.e. drop
  # the `target` matrix), and maybe without some configurations (i.e. no threads and
  # serial?)
```

The content of the files `as_subdirectory.sh` and `as_library.sh` should appear within the test as maybe two separate steps. Having script files lying around in the source tree that are only used for CI is not a good thing. Instead of re-downloading Kokkos, I think we can re-use the one in `/tpls` by copying it (which makes it consistent with the build tests).

I don't like the idea to duplicate the install test file, but I think it's better to keep two separate directories for the library and the subdirectory approaches, as they require different `CMakeLists.txt` files. Maybe just reduce the test file to its simplest expression.

yasahi-hpc:
@pzehner Thanks for the comments

> Based on our previous discussion, testing the installation of KokkosFFT should be done in a different job:
> 
> ```yaml
> install:
>   # same `needs` and `if` as the `build` job.
>   # the `strategy` may be similar, without testing host-device configuration (i.e. drop
>   # the `target` matrix), and maybe without some configurations (i.e. no threads and
>   # serial?)
> ```
> 
> The content of the files `as_subdirectory.sh` and `as_library.sh` should appear within the test as maybe two separate steps. Having script files lying around in the source tree that are only used for CI is not a good thing. Instead of re-downloading Kokkos, I think we can re-use the one in `/tpls` by copying it (which makes it consistent with the build tests).

That makes sense. I will update accordingly. 

> I don't like the idea to duplicate the install test file, but I think it's better to keep two separate directories for the library and the subdirectory approaches, as they require different `CMakeLists.txt` files. Maybe just reduce the test file to its simplest expression.

Do you mean we should modify each `CMakeLists.txt` to refer to a single source file, say, 

```CMake
add_executable(hello-kokkos-fft ../src/hello.cpp)
```

yasahi-hpc:
> > Do you mean we should modify each `CMakeLists.txt` to refer to a single source file, say,
> > ```cmake
> > add_executable(hello-kokkos-fft ../src/hello.cpp)
> > ```
> 
> We would end up with the following layout:
> 
> ```
> install_test
>     +-- main.cpp
>     +-- as_library
>     |   +-- CMakeLists.txt
>     +-- as_subdirectory
>         +-- CMakeLists.txt
> ```
> 
> Which is not very common. It's up to you I guess.

I may prefer self-contained examples like

```
 install_test
    +-- as_library
    |   +-- CMakeLists.txt     
    |   +-- hello.cpp
    +-- as_subdirectory
        +-- CMakeLists.txt
        +-- hello_again.cpp
```

Using slightly different examples. Is it fine with you?

install_test/as_subdirectory/as_subdirectory.sh
```diff
@@ -0,0 +1,75 @@
...
+cp ${SRC_DIR}/install_test/as_subdirectory/* .
+
+# Prepare Kokkos and KokkosFFT under tpls
+mkdir tpls && cd tpls
+
+# Prepare tpls/kokkos 
+git clone https://github.com/kokkos/kokkos.git
+
+# Copy KokkosFFT under tpls/kokkos-fft
+cp -r ${SRC_DIR} kokkos-fft
```
cedricchevalier19, line 55: Looks very complicated. Why not just go to this directory and have a CMakeLists.txt that looks like https://github.com/kokkos/kokkos/blob/e0dc0128e04f18c2bbbaefceef3616e7ddcfa3c4/example/build_cmake_in_tree/CMakeLists.txt#L10

yasahi-hpc: I can try, but not sure it works. Using Kokkos added as a submodule causes an error in the git check. https://github.com/CExA-project/kokkos-fft/actions/runs/8008651458/job/21875544621

In addition, I would like a CMakeLists.txt and build commands exactly the same as used in the docs.

pzehner: This file was finally removed.

yasahi-hpc: Great!

PR #73, opened 2024-02-29
Update docs

yasahi-hpc:
This PR aims at final updating of docs before release.

1. files under `docs` updated based on recent changes
2. README.md updated
3. fix issues in Doxygen

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,49 +108,7 @@ cmake -DCMAKE_CXX_COMPILER=g++ \
       -DKokkos_ENABLE_CUDA=ON \
       -DKokkos_ARCH_AMPERE80=ON ..
 ```
-This way, all the functionalities are executed on A100 GPUs.
-
-### Install as a library
-Is is assumed that the Kokkos is installed under `<install_dir>/kokkos` with OpenMP backend. Here is a recipe to install KokkosFFT under `<install_dir>/kokkosFFT`.
-
-```bash
-export KOKKOSFFT_INSTALL_PREFIX=<lib_dir>/kokkosFFT
-
-mkdir build_KokkosFFT && cd build_KokkosFFT
-cmake -DCMAKE_CXX_COMPILER=icpx \
-      -DCMAKE_INSTALL_PREFIX=${KOKKOSFFT_INSTALL_PREFIX} ..
-cmake --build . -j 8
-cmake --install .
-```
-
-Here is an example to use KokkosFFT in the following CMake project.
-```
----/
- |
- └──<project_directory>/
-    |--CMakeLists.txt
-    └──hello.cpp
-```
-
-The `CMakeLists.txt` would be
-```CMake
-cmake_minimum_required(VERSION 3.23)
-project(kokkos-fft-as-library LANGUAGES CXX)
-
-find_package(Kokkos CONFIG REQUIRED)
-find_package(KokkosFFT CONFIG REQUIRED)
-
-add_executable(hello-kokkos-fft hello.cpp)
-target_link_libraries(hello-kokkos-fft PUBLIC Kokkos::kokkos KokkosFFT::fft)
-```
-
-The code can be built as
-```bash
-mkdir build && cd build
-cmake -DCMAKE_CXX_COMPILER=icpx \
-      -DCMAKE_PREFIX_PATH="<install_dir>/kokkos;<install_dir>/kokkosFFT" ..
-cmake --build . -j 8
-```
+This way, all the functionalities are executed on A100 GPUs. Further installation details are provided in the [documentation](./docs/intro/building.rst).
 
 ## LICENCE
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -111,11 +111,10 @@ cmake -DCMAKE_CXX_COMPILER=g++ \
 This way, all the functionalities are executed on A100 GPUs.
```
pzehner, line 111: I think it's not the best idea to duplicate install instructions on both the readme and on the doc. This is an error-prone process (DRY). Since GitHub can display RST files, a different approach is possible.

You should have a condensed section on how to use KokkosFFT as a subdirectory dependency (this is likely how a programmer would start using this library), then have an anchor to the documentation where the complete install process is given:

```md
Further installation details are provided in the [documentation](./docs/intro/building.rst).
```

I think this approach could be used for other sections of the readme. In other words, this file (which is the first contact for a potential user) would be a simple starter, redirecting to more complete documentation pages.

docs/intro/building.rst
```diff
@@ -63,7 +63,8 @@ The code can be built as
 .. code-block:: bash
 
     mkdir build && cd build
-    cmake -DCMAKE_PREFIX_PATH="<install_dir>/kokkos;<install_dir>/kokkos_fft" ..
+    cmake -DCMAKE_CXX_COMPILER=icpx \
```
pzehner, line 66: Why advertise for the Intel compiler?

You could be more generic with `-DCMAKE_CXX_COMPILER=<your C++ compiler>`.

docs/intro/building.rst
```diff
@@ -22,7 +22,7 @@ However, we have not tested all the listed compilers there and thus recommend th
 Install KokkosFFT as a library
 ------------------------------
 
-Is is assumed that Kokkos is installed under ``<install_dir>/kokkos`` with OpenMP backend. Here is a recipe to install KokkosFFT under ``<install_dir>/kokkos_fft``.
+Is is assumed that Kokkos is installed under ``<install_dir>/kokkos`` with OpenMP backend. Here is a recipe to install KokkosFFT under ``<install_dir>/kokkosFFT``.
```
pzehner, line 25: I think you can rephrase this as:

```rst
Let's assume Kokkos is installed under ``path/to/kokkos`` with OpenMP backend. We build and install KokkosFFT under ``path/to/kokkos-fft``.
```

docs/intro/building.rst
```diff
@@ -63,7 +63,8 @@ The code can be built as
 .. code-block:: bash
 
     mkdir build && cd build
-    cmake -DCMAKE_PREFIX_PATH="<install_dir>/kokkos;<install_dir>/kokkos_fft" ..
+    cmake -DCMAKE_CXX_COMPILER=icpx \
+          -DCMAKE_PREFIX_PATH="<install_dir>/kokkos;<install_dir>/kokkosFFT" ..
```
pzehner, line 67: This line would become

```
    -DKMAKE_PREFIX_PATH=path/to/kokkos;path/to/kokkos-fft`
```

Alternatively, you can use the syntax:

```
    -DKokkos_ROOT=path/to/kokkos
    -DKokkosFFT_ROOT=path/to/kokkos-fft
```

The latest version of the cheat sheet tends to use this way.

pzehner (changes requested):
This is a nice update, but I think that information in the documentation shouldn't be duplicated in the readme.

docs/intro/building.rst
yasahi-hpc: Then `g++`? I would like to make it as easy as possible, i.e., ideally, copy and paste this part is sufficient.

docs/intro/building.rst
yasahi-hpc: Sure

docs/intro/building.rst
yasahi-hpc: In the future release, we may use 

```
    -DKokkos_ROOT=path/to/kokkos
    -DKokkosFFT_ROOT=path/to/kokkos-fft
```


For the moment, we use the `-DKMAKE_PREFIX_PATH` version in CI, so I would keep in the docs as well

README.md
yasahi-hpc: It is indeed a duplicate if people read both `README.md` and docs. However, people may not necessarily read the docs. Not sure what is the right way.

docs/intro/building.rst
pzehner: Then don't put anything. CMake would check the environment variable `CC`, and ultimately default to GCC.

docs/intro/building.rst
pzehner: Ok.

docs/intro/building.rst
yasahi-hpc: I fixed accordingly. I will merge this and make a release candidate. Thank you for your review

yasahi-hpc:
Close this PR, because of asyucrhouns commits issue.

PR #91, opened 2024-03-15
Cleanup plan class

### LINKED ISSUE #90: Cleanup Plan class. Disallow copies.

yasahi-hpc:
Resolves #90 

In this PR, we
1. disabled the non-default constructors of `Plan` class
2. added some tests for reshaping parameter `s`

### DIFF
--- a/fft/src/KokkosFFT_Plans.hpp
+++ b/fft/src/KokkosFFT_Plans.hpp
@@ -97,9 +97,6 @@ class Plan {
   //! The type of extents of input/output views
   using extents_type = shape_type<InViewType::rank()>;
 
-  //! Execution space
-  execSpace m_exec_space;
-
   //! Dynamically allocatable fft plan.
   std::unique_ptr<fft_plan_type> m_plan;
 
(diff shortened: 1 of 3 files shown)


fft/src/KokkosFFT_Plans.hpp
```diff
@@ -97,6 +97,9 @@ class Plan {
   //! The type of extents of input/output views
   using extents_type = shape_type<InViewType::rank()>;
 
+  //! Execution space
+  execSpace m_exec_space;
```
dalg24, line 101: Why do you hold on to an exec space?
When one of the API function is called presumably you take an exec space as argument and you pass the plan.  Then which exec space do you pick?  Do they have to match?  etc..

dalg24 (commented):
(more a drive-by comment than a thorough review)
What is the exec space argument to the Plan constructor intended for?  Allocation of internal buffers?  What else?  Why is it ok to hold on to that exec space?  Should it be reused later when actually computing the FFTs?

fft/src/KokkosFFT_Plans.hpp
yasahi-hpc: For the moment, this internal `exec_space` is unused. I will suppress this change in this PR. 
Originally, I was inclined to use internal `exec_space` for all the execution of FFTs and Kokkos functions, but it turns out that the changes are too big to be consistent.

yasahi-hpc: Thank you for your review. I will merge this. Your reviews and/or suggestions are always appreciated!

PR #92, opened 2024-03-19
Enable SYCL testing on PVC2

pzehner:
This PR aims to test KokkosFFT built against the SYCL backend on an Intel GPU Max 1550.


pzehner:
Closed as the access to this machine was lost.

Issue #103, opened 2024-06-18
Recover GPU runner


yasahi-hpc:
Self-hosted runner needs to be recovered as we have not used it for 30 days. (See https://github.com/actions/runner/issues/756)

PR #104, opened 2024-06-19
Setup CI heartbeat for self-hosted runners

pzehner:
This PR uses a simple way to prevent any self-hosted runner to shutdown if not used for 30 consecutive days (as discussed in [this issue](https://github.com/actions/runner/issues/756)). It schedules a dummy job to execute on this runner twice a month


yasahi-hpc:
Resolves #103

Issue #106, opened 2024-06-26
Stricter checks in the CI


cedricchevalier19:
Ideally we should not allow warnings.
I am in favor to add `CMAKE_CXX_FLAGS="-Wall -Werror"` to the CI.

yasahi-hpc:
Makes sense. I need to investigate warnings from each compiler.

PR #109, opened 2024-07-03
Cleaning up some warnings and enabling -Werror on AMD GPU CI tests

yasahi-hpc:
Improve https://github.com/kokkos/kokkos-fft/issues/106.

- Cleaning hipcc warnings on AMD GPUs
- Enabling warnings and warning as error (`-Werror`)

### DIFF
--- a/fft/src/KokkosFFT_OpenMP_plans.hpp
+++ b/fft/src/KokkosFFT_OpenMP_plans.hpp
@@ -12,7 +12,7 @@
 namespace KokkosFFT {
 namespace Impl {
 template <typename ExecutionSpace, typename T>
-void _init_threads(const ExecutionSpace& exec_space) {
+void _init_threads([[maybe_unused]] const ExecutionSpace& exec_space) {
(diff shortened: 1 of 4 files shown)


fft/src/KokkosFFT_OpenMP_plans.hpp
```diff
@@ -12,7 +12,7 @@
 namespace KokkosFFT {
 namespace Impl {
 template <typename ExecutionSpace, typename T>
-void _init_threads(const ExecutionSpace& exec_space) {
+void _init_threads([[maybe_unused]] const ExecutionSpace& exec_space) {
```
cedricchevalier19, line 15: Perhaps this file should be renamed CPU or host?

cedricchevalier19 (approved):

fft/src/KokkosFFT_OpenMP_plans.hpp
yasahi-hpc: Makes sense. Maybe better to call it Host? I will fix this in another PR.

PR #112, opened 2024-07-04
Cleaning up some warnings and enabling -Werror on Intel GPU CI tests

yasahi-hpc:
Resolves https://github.com/kokkos/kokkos-fft/issues/106.

- Cleaning icpx warnings on Intel GPUs
- Enabling warnings, but cannot use `-Werror` because there are already many warnings from Kokkos


cedricchevalier19 (approved):
An external Kokkos build may avoid warnings from Kokkos (by using `-isystem`).

Anyway it is a great improvement.

yasahi-hpc:
> An external Kokkos build may avoid warnings from Kokkos (by using `-isystem`).
> 
> Anyway it is a great improvement.

Thank you for your review.
So we need to modify CMake a little bit?

The easiest way may be to do the build test with `-Werror` option using the preinstalled Kokkos
I will do that in another PR

PR #114, opened 2024-07-05
Rename: KokkosFFT_OpenMP*.hpp to KokkosFFT_Host*.hpp for fftw backend

### LINKED ISSUE #110: Rename KokkosFFT_OpenMP_*.hpp to KokkosFFT_Host_*.hpp
KokkosFFT_OpenMP_*.hpp handles `OpenMP`, `Threads` and `Serial`.
It is better to rename `Host` as suggested in #109

yasahi-hpc:
Resolves #110 

Renaming files for fftw backend:
```
 rename fft/src/{KokkosFFT_OpenMP_plans.hpp => KokkosFFT_Host_plans.hpp} (98%)
 rename fft/src/{KokkosFFT_OpenMP_transform.hpp => KokkosFFT_Host_transform.hpp} (94%)
 rename fft/src/{KokkosFFT_OpenMP_types.hpp => KokkosFFT_Host_types.hpp} (97%)
```

### DIFF
--- a/fft/src/KokkosFFT_Host_plans.hpp
+++ b/fft/src/KokkosFFT_Host_plans.hpp
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
 
-#ifndef KOKKOSFFT_OPENMP_PLANS_HPP
-#define KOKKOSFFT_OPENMP_PLANS_HPP
+#ifndef KOKKOSFFT_HOST_PLANS_HPP
+#define KOKKOSFFT_HOST_PLANS_HPP
 
 #include <numeric>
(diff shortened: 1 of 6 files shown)


jbigot:
is it host or parallel-host? By that I mean: is there a separate set of files for host serial?

cedricchevalier19:
@jbigot, I think it is from #109; I saw that the `OpenMP` files are also for `Threads` and `Serial` backends.

jbigot (approved):

yasahi-hpc:
@jbigot

> is it host or parallel-host? By that I mean: is there a separate set of files for host serial?

As @cedricchevalier19 said, this handles all host backends, so `Host` should be a better name.
In the current implementation, the only difference between `parallel_host` and `host`  is the call of `init_threads` for `fftw` or not

Issue #80, opened 2024-03-11
Error handling let them know which line causes the error


yasahi-hpc:

yasahi-hpc:
Static assertion to Unary (`fftshift`, `ifftshift`) and Binary (`Plan` and `fft` functions) APIs.
[ ] Exec space and memory space consistency (all)
[ ] precision consistency (Binary)
[ ] layout consistency (Binary)
[ ] rank consistency (Binary)
[ ] Introduce `is_view_operatable_v` for Unary operation
[ ] Introduce `are_views_operatable_v` for Binary operation
[ ] Introduce `have_same_precision_v` for Binary operation
[ ] Introduce `have_same_layout_v` for Binary operation
[ ] Introduce `have_same_rank_v` for Binary operation
  
Dynamic assertion should be `std::runtime_error` with appropriate messages.
If there is inconsistency in extents, we cannot operate FFT.
[ ] replace `assert` with `std::runtime_error`

PR #119, opened 2024-07-15
More traits for unary and binary operations

yasahi-hpc:
This PR aims at adding more traits for unary and binary operations to `KokkosFFT_traits.hpp`.
This will be applied in another PR to improve error handlings as commented in #80

Static assertion to Unary (`fftshift`, `ifftshift`) and Binary (`Plan` and `fft` functions) APIs. Applications of new traits (unchecked tasks) would be done in another PR.

- [x] Introduce `is_view_operatable_v` for Unary operation (Views can be accessible from `Execspace` and the base value types are `float/double/Kokkos::complex<float>/Kokkos::complex<double>`. In addition, layout is either `Kokkos::LayoutLeft` or `Kokkos::LayoutRight`.
- [x] Introduce `are_views_operatable_v` for Binary operation (Both Views are `operatable` and have the same `precision`, `layout` and `rank`)
- [x] Introduce ~~`have_same_precision_v`~~ `have_same_base_floating_point_type_v`  for Binary operation (the values have the same ~~precision~~ base floating point type. For example, `float` and `Kokkos::complex<float>` have the same precision)
- [x] Introduce `have_same_layout_v` for Binary operation
- [x] Introduce `have_same_rank_v` for Binary operation
- [ ] Apply `are_views_operatable_v` to binary APIs. Exec space and memory space consistency, precision consistency, layout consistency and rank consistency.
- [ ] Apply `is_view_operatable_v` to unary APIs. Exec space and memory space consistency.

### DIFF
--- a/common/src/KokkosFFT_traits.hpp
+++ b/common/src/KokkosFFT_traits.hpp
@@ -131,10 +131,10 @@ inline constexpr bool is_operatable_view_v =
 
 // Traits for binary operations
 template <typename T1, typename T2, typename Enable = void>
-struct have_same_precision : std::false_type {};
+struct have_same_base_floating_point_type : std::false_type {};
 
 template <typename T1, typename T2>
-struct have_same_precision<
+struct have_same_base_floating_point_type<
     T1, T2,
     std::enable_if_t<!Kokkos::is_view_v<T1> && !Kokkos::is_view_v<T2> &&
                      std::is_same_v<base_floating_point_type<T1>,
@@ -142,7 +142,7 @@ struct have_same_precision<
     : std::true_type {};
 
 template <typename InViewType, typename OutViewType>
-struct have_same_precision<
+struct have_same_base_floating_point_type<
     InViewType, OutViewType,
     std::enable_if_t<
         Kokkos::is_view_v<InViewType> && Kokkos::is_view_v<OutViewType> &&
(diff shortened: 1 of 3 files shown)


common/src/KokkosFFT_traits.hpp
```diff
@@ -107,6 +110,111 @@ template <typename ViewType>
...
+template <typename T1, typename T2>
+struct have_same_precision<
+    T1, T2,
+    std::enable_if_t<!Kokkos::is_view_v<T1> && !Kokkos::is_view_v<T2> &&
+                     std::is_same_v<base_floating_point_type<T1>,
+                                    base_floating_point_type<T2>>>>
+    : std::true_type {};
+
+template <typename InViewType, typename OutViewType>
+struct have_same_precision<
```
cedricchevalier19, line 145: Not sure if "precision" is the correct name. Is it the same base floating point type?

common/unit_test/Test_Traits.cpp
```diff
@@ -18,21 +28,74 @@ using view_types =
...
+    tuple_to_types_t<cartesian_product_t<base_real_types, base_layout_types,
+                                         base_real_types, base_layout_types>>;
+
 template <typename T>
 struct RealAndComplexTypes : public ::testing::Test {
   using real_type    = T;
   using complex_type = Kokkos::complex<T>;
+
+  virtual void SetUp() {
+    GTEST_SKIP() << "Skipping all tests for this fixture";
```
cedricchevalier19, line 48: Do you mean that tests are only compilation and therefore there is no runtime test?
If it is the case, I think using gtest is quite a burden but perhaps the skip explanation can be more straightforward.

cedricchevalier19 (commented):
Looks ok, but I am not sure about the PR description, a lot of items are not checked. I think it is ok to merge if you plan to put the other parts in a separate PR.

common/src/KokkosFFT_traits.hpp
yasahi-hpc: Sure, I will change the name accordingly.

common/unit_test/Test_Traits.cpp
yasahi-hpc: Yes, this is a compilation test.
I have commented this at the top of this file. 
Maybe better to comment that `gtest` is used for parameterization only?

yasahi-hpc:
> Looks ok, but I am not sure about the PR description, a lot of items are not checked. I think it is ok to merge if you plan to put the other parts in a separate PR.

Sorry description was unclear. I have updated the description. 
As you said, this PR aims to introduce new traits. These traits are then applied in a separate PR.

cedricchevalier19 (approved):

PR #122, opened 2024-07-24
Replace assert with runtime error

yasahi-hpc:
Improves #80

This PR aims at replacing runtime `assert` with `std::runtime_error` with appropriate messages.
If there is inconsistency in extents, we cannot operate FFT.

Modifications are applied to `get_shift`, `get_extents`, `get_modified_shape` and `convert_negative_axis`. For example, in `get_shift`, we replaced `assert` with `std::runtime_error` in the following way.

```C++
// Previous check
assert(!KokkosFFT::Impl::has_duplicate_values(axes));

// Current check
//if (KokkosFFT::Impl::has_duplicate_values(axes)) {
//    throw std::runtime_error("get_shift: axes are overlapped.");
//}
//check_precondition(!KokkosFFT::Impl::has_duplicate_values(axes),
//                   "axes are overlapped");
KOKKOSFFT_EXPECTS(!KokkosFFT::Impl::has_duplicate_values(axes),
                  "axes are overlapped");
```

- [x] Replacing runtime `assert` with `std::runtime_error`
- [x] Add corresponding tests to check `std::runtime_error` is correctly thrown.
- [x] Add missing test for `has_duplicate_values`
- [x] Introduce ~~`check_precondition(expression, message)` function~~ `KOKKOSFFT_EXPECTS(expression, message)`macro

### DIFF
--- a/common/src/KokkosFFT_Helpers.hpp
+++ b/common/src/KokkosFFT_Helpers.hpp
@@ -23,12 +23,12 @@ auto get_shift(const ViewType& inout, axis_type<DIM> _axes, int direction = 1) {
 
   // Assert if the elements are overlapped
   constexpr int rank = ViewType::rank();
-  check_precondition(!KokkosFFT::Impl::has_duplicate_values(axes),
-                     "axes are overlapped");
-  check_precondition(
+  KOKKOSFFT_EXPECTS(!KokkosFFT::Impl::has_duplicate_values(axes),
+                    "Axes overlap");
+  KOKKOSFFT_EXPECTS(
       !KokkosFFT::Impl::is_out_of_range_value_included(axes, rank),
-      "axes include out of range index."
-      "axes should be in the range of [-rank, rank-1].");
+      "Axes include an out-of-range index."
+      "Axes must be in the range of [-rank, rank-1].");
 
   axis_type<rank> shift = {0};
   for (int i = 0; i < static_cast<int>(DIM); i++) {
--- a/common/src/KokkosFFT_Helpers.hpp
+++ b/common/src/KokkosFFT_Helpers.hpp
@@ -23,8 +23,14 @@ auto get_shift(const ViewType& inout, axis_type<DIM> _axes, int direction = 1) {
 
   // Assert if the elements are overlapped
   constexpr int rank = ViewType::rank();
-  assert(!KokkosFFT::Impl::has_duplicate_values(axes));
-  assert(!KokkosFFT::Impl::is_out_of_range_value_included(axes, rank));
+  if (KokkosFFT::Impl::has_duplicate_values(axes)) {
+    throw std::runtime_error("get_shift: axes are overlapped.");
+  }
+  if (KokkosFFT::Impl::is_out_of_range_value_included(axes, rank)) {
+    throw std::runtime_error(
+        "get_shift: axes include out of range index."
+        "axes should be in the range of [-rank, rank-1].");
+  }
 
   axis_type<rank> shift = {0};
   for (int i = 0; i < static_cast<int>(DIM); i++) {
--- a/common/src/KokkosFFT_Helpers.hpp
+++ b/common/src/KokkosFFT_Helpers.hpp
@@ -23,14 +23,12 @@ auto get_shift(const ViewType& inout, axis_type<DIM> _axes, int direction = 1) {
 
   // Assert if the elements are overlapped
   constexpr int rank = ViewType::rank();
-  if (KokkosFFT::Impl::has_duplicate_values(axes)) {
-    throw std::runtime_error("get_shift: axes are overlapped.");
-  }
-  if (KokkosFFT::Impl::is_out_of_range_value_included(axes, rank)) {
-    throw std::runtime_error(
-        "get_shift: axes include out of range index."
-        "axes should be in the range of [-rank, rank-1].");
-  }
+  check_precondition(!KokkosFFT::Impl::has_duplicate_values(axes),
+                     "axes are overlapped");
+  check_precondition(
+      !KokkosFFT::Impl::is_out_of_range_value_included(axes, rank),
+      "axes include out of range index."
+      "axes should be in the range of [-rank, rank-1].");
 
   axis_type<rank> shift = {0};
   for (int i = 0; i < static_cast<int>(DIM); i++) {
--- a/common/src/KokkosFFT_layouts.hpp
+++ b/common/src/KokkosFFT_layouts.hpp
@@ -67,12 +67,11 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   if (is_real_v<in_value_type>) {
     // Then R2C
     if (is_complex_v<out_value_type>) {
-      if (_out_extents.at(inner_most_axis) !=
-          _in_extents.at(inner_most_axis) / 2 + 1) {
-        throw std::runtime_error(
-            "For R2C, the output extent of transform should be input extent / "
-            "2 + 1");
-      }
+      KOKKOSFFT_EXPECTS(
+          _out_extents.at(inner_most_axis) ==
+              _in_extents.at(inner_most_axis) / 2 + 1,
+          "For R2C, the 'output extent' of transform must be equal to "
+          "'input extent'/2 + 1");
     } else {
       throw std::runtime_error(
           "If the input type is real, the output type should be complex");
@@ -82,12 +81,11 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   if (is_real_v<out_value_type>) {
     // Then C2R
     if (is_complex_v<in_value_type>) {
-      if (_in_extents.at(inner_most_axis) !=
-          _out_extents.at(inner_most_axis) / 2 + 1) {
-        throw std::runtime_error(
-            "For C2R, the input extent of transform should be output extent / "
-            "2 + 1");
-      }
+      KOKKOSFFT_EXPECTS(
+          _in_extents.at(inner_most_axis) ==
+              _out_extents.at(inner_most_axis) / 2 + 1,
+          "For C2R, the 'input extent' of transform must be equal to "
+          "'output extent' / 2 + 1");
     } else {
       throw std::runtime_error(
           "If the output type is real, the input type should be complex");
--- a/common/src/KokkosFFT_layouts.hpp
+++ b/common/src/KokkosFFT_layouts.hpp
@@ -67,8 +67,12 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   if (is_real_v<in_value_type>) {
     // Then R2C
     if (is_complex_v<out_value_type>) {
-      assert(_out_extents.at(inner_most_axis) ==
-             _in_extents.at(inner_most_axis) / 2 + 1);
+      if (_out_extents.at(inner_most_axis) !=
+          _in_extents.at(inner_most_axis) / 2 + 1) {
+        throw std::runtime_error(
+            "For R2C, the output extent of transform should be input extent / "
+            "2 + 1");
+      }
     } else {
       throw std::runtime_error(
           "If the input type is real, the output type should be complex");
@@ -78,8 +82,12 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   if (is_real_v<out_value_type>) {
     // Then C2R
     if (is_complex_v<in_value_type>) {
-      assert(_in_extents.at(inner_most_axis) ==
-             _out_extents.at(inner_most_axis) / 2 + 1);
+      if (_in_extents.at(inner_most_axis) !=
+          _out_extents.at(inner_most_axis) / 2 + 1) {
+        throw std::runtime_error(
+            "For C2R, the input extent of transform should be output extent / "
+            "2 + 1");
+      }
     } else {
       throw std::runtime_error(
           "If the output type is real, the input type should be complex");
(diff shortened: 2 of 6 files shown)


dalg24:
Maybe you could abstract that away and have a precondition check utility that takes a message as argument.

yasahi-hpc:
> Maybe you could abstract that away and have a precondition check utility that takes a message as argument.

Makes sense. So I would introduce something like this?
I will do that in another PR

```C++

void throw_runtime_error(const bool condition, const std::string &msg) {
  if (condition) {
    throw std::runtime_error(msg);
  }
}
```

dalg24:
You probably want to call it something else, possibly `check_precondition(expression, message)`.
If your minimum requirement allows it, consider having a std::source_location trailing argument defaulted to "current".  Otherwise you could look into using a macro.

yasahi-hpc:
> You probably want to call it something else, possibly `check_precondition(expression, message)`. If your minimum requirement allows it, consider having a std::source_location trailing argument defaulted to "current". Otherwise you could look into using a macro.

That is a better name. We are currently supporting C++17, so maybe using a macro. 
I am also interested in introducing C++20 features, so I will give it a try with a protection by `KOKKOS_ENABLE_CXX17` macro

cedricchevalier19:
@yasahi-hpc: do you have strong requirements from users for supporting C++17 and not requiring C++20?
I think if we can avoid `#ifdef` it will be a lot better.

yasahi-hpc:
> @yasahi-hpc: do you have strong requirements from users for supporting C++17 and not requiring C++20? I think if we can avoid `#ifdef` it will be a lot better.

I do not think so. I was just planning to require C++20 at the same time as Kokkos.

yasahi-hpc:
@cedricchevalier19 
For the moment, I am relying on macros. Is this OK for you?

common/src/KokkosFFT_utils.hpp
```diff
@@ -8,19 +8,45 @@
...
+  std::stringstream ss("file: ");
+#ifndef KOKKOS_ENABLE_CXX17
+  std::source_location location = std::source_location::current();
+  ss << location.file_name() << '(' << location.line() << ':'
+     << location.column() << ") `" << location.function_name() << "`: " << msg
+     << '\n';
+#else
+  ss << __FILE__ << '(' << __LINE__ << ") `" << __FUNCTION__ << "`: " << msg
+     << '\n';
+#endif
```
dalg24, line 35: This is always going to point to this file, not the calling site.  That is why I had suggested making the actual assertion a macro.

yasahi-hpc: Sorry, I thought they are equivalent. Fixed now

common/src/KokkosFFT_utils.hpp
```diff
@@ -8,19 +8,59 @@
...
 #include <set>
+#include <string>
 #include <algorithm>
 #include <numeric>
 #include "KokkosFFT_traits.hpp"
 
+#ifndef KOKKOS_ENABLE_CXX17
+#include <source_location>
+#define KOKKOSFFT_EXPECTS(expression, msg) \
+  KokkosFFT::Impl::check_precondition(expression, msg)
```
dalg24, line 19:
```suggestion
  KokkosFFT::Impl::check_precondition(expression, msg, std::source_location::current())
```

common/src/KokkosFFT_utils.hpp
```diff
@@ -8,19 +8,59 @@
...
+#define KOKKOSFFT_EXPECTS(expression, msg)                                 \
+  KokkosFFT::Impl::check_precondition(expression, msg, __FILE__, __LINE__, \
+                                      __FUNCTION__)
+#endif
+
 namespace KokkosFFT {
 namespace Impl {
 
+#ifndef KOKKOS_ENABLE_CXX17
+inline void check_precondition(const bool expression, const std::string& msg) {
```
dalg24, line 31:
```suggestion
inline void check_precondition(const bool expression, const std::string& msg, std::source_location location) {
```

common/src/KokkosFFT_utils.hpp
```diff
@@ -8,19 +8,59 @@
...
+  std::stringstream ss("file: ");
+  std::source_location location = std::source_location::current();
```
dalg24, line 33:
```suggestion
```

yasahi-hpc:
@dalg24 I guess is OK now. Can I have your second review please

common/src/KokkosFFT_utils.hpp
```diff
@@ -8,19 +8,60 @@
...
 
+#ifndef KOKKOS_ENABLE_CXX17
+#include <source_location>
+#define KOKKOSFFT_EXPECTS(expression, msg)             \
+  KokkosFFT::Impl::check_precondition(expression, msg, \
+                                      std::source_location::current())
+#else
+#include <cstdlib>
+#define KOKKOSFFT_EXPECTS(expression, msg)                                 \
+  KokkosFFT::Impl::check_precondition(expression, msg, __FILE__, __LINE__, \
```
dalg24, line 24: I suspect you need the extraneous parentheses
```suggestion
  KokkosFFT::Impl::check_precondition((expression), msg, __FILE__, __LINE__, \
```

dalg24 (commented):
I haven't looked in details but it looks like the right direction

common/src/KokkosFFT_utils.hpp
yasahi-hpc: Fixed accordingly

yasahi-hpc:
> I haven't looked in details but it looks like the right direction

I think so. I am inclined to improve the compile time and runtime errors. Still insufficient though

common/src/KokkosFFT_Helpers.hpp
```diff
@@ -23,8 +23,12 @@ auto get_shift(const ViewType& inout, axis_type<DIM> _axes, int direction = 1) {
 
   // Assert if the elements are overlapped
   constexpr int rank = ViewType::rank();
-  assert(!KokkosFFT::Impl::has_duplicate_values(axes));
-  assert(!KokkosFFT::Impl::is_out_of_range_value_included(axes, rank));
+  KOKKOSFFT_EXPECTS(!KokkosFFT::Impl::has_duplicate_values(axes),
+                    "axes are overlapped");
```
cedricchevalier19, line 27:
```suggestion
  KOKKOSFFT_EXPECTS(!KokkosFFT::Impl::has_duplicate_values(axes),
                    "Axes overlap");
```

common/src/KokkosFFT_Helpers.hpp
```diff
@@ -23,8 +23,12 @@ auto get_shift(const ViewType& inout, axis_type<DIM> _axes, int direction = 1) {
...
+  KOKKOSFFT_EXPECTS(
+      !KokkosFFT::Impl::is_out_of_range_value_included(axes, rank),
+      "axes include out of range index."
+      "axes should be in the range of [-rank, rank-1].");
```
cedricchevalier19, line 31:
```suggestion
  KOKKOSFFT_EXPECTS(
      !KokkosFFT::Impl::is_out_of_range_value_included(axes, rank),
      "Axes include an out-of-range index."
      "Axes must be in the range of [-rank, rank-1].");
```

common/src/KokkosFFT_layouts.hpp
```diff
@@ -67,8 +67,11 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   if (is_real_v<in_value_type>) {
     // Then R2C
     if (is_complex_v<out_value_type>) {
-      assert(_out_extents.at(inner_most_axis) ==
-             _in_extents.at(inner_most_axis) / 2 + 1);
+      KOKKOSFFT_EXPECTS(
+          _out_extents.at(inner_most_axis) ==
+              _in_extents.at(inner_most_axis) / 2 + 1,
+          "For R2C, the output extent of transform should be input extent / "
+          "2 + 1");
```
cedricchevalier19, line 74:
```suggestion
          "For R2C, the 'output extent' of transform must be equal to "
          "'input extent'/2 + 1");
```

common/src/KokkosFFT_layouts.hpp
```diff
@@ -78,8 +81,11 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   if (is_real_v<out_value_type>) {
     // Then C2R
     if (is_complex_v<in_value_type>) {
-      assert(_in_extents.at(inner_most_axis) ==
-             _out_extents.at(inner_most_axis) / 2 + 1);
+      KOKKOSFFT_EXPECTS(
+          _in_extents.at(inner_most_axis) ==
+              _out_extents.at(inner_most_axis) / 2 + 1,
+          "For C2R, the input extent of transform should be output extent / "
+          "2 + 1");
```
cedricchevalier19, line 88:
```suggestion
          "For C2R, the 'input extent' of transform must be equal to "
          "'output extent'/2 + 1");
```

common/src/KokkosFFT_padding.hpp
```diff
@@ -51,8 +51,12 @@ auto get_modified_shape(const InViewType in, const OutViewType /* out */,
   }
 
   // Assert if the elements are overlapped
-  assert(!KokkosFFT::Impl::has_duplicate_values(positive_axes));
-  assert(!KokkosFFT::Impl::is_out_of_range_value_included(positive_axes, rank));
+  KOKKOSFFT_EXPECTS(!KokkosFFT::Impl::has_duplicate_values(positive_axes),
+                    "axes are overlapped");
```
cedricchevalier19, line 55:
```suggestion
                    "Axes overlap");
```

common/src/KokkosFFT_padding.hpp
```diff
@@ -51,8 +51,12 @@ auto get_modified_shape(const InViewType in, const OutViewType /* out */,
...
+  KOKKOSFFT_EXPECTS(
+      !KokkosFFT::Impl::is_out_of_range_value_included(positive_axes, rank),
+      "axes include out of range index."
+      "axes should be in the range of [-rank, rank-1].");
```
cedricchevalier19, line 59:
```suggestion
      "Axes include an out-of-range index."
      "Axes must be in the range of [-rank, rank-1].");
```

common/src/KokkosFFT_utils.hpp
```diff
@@ -8,19 +8,60 @@
...
+
 template <typename ViewType>
 auto convert_negative_axis(ViewType, int _axis = -1) {
   static_assert(Kokkos::is_view<ViewType>::value,
                 "convert_negative_axis: ViewType is not a Kokkos::View.");
   int rank = static_cast<int>(ViewType::rank());
-  assert(_axis >= -rank && _axis < rank);  // axis should be in [-rank, rank-1]
+
+  KOKKOSFFT_EXPECTS(_axis >= -rank && _axis < rank,
+                    "axis should be in [-rank, rank-1]");
```
cedricchevalier19, line 63:
```suggestion
                    "axis must be in [-rank, rank-1]");
```

common/src/KokkosFFT_utils.hpp
cedricchevalier19, line 1: For this file, I think we can:
- Do things more straightforward, having `#ifdef` in `check_precondition` core or even having the code directly in the macro
- Or, let the `Impl::check_precondition` be implemented only once (with the 5 arguments prototype, you can use `std::source_location` in your macro. Ideally, the implementation can be in a separate C++ file to avoid exposing dependencies to `sstream` (where is it included, by the way?).
- Only support EXPECTS for C++20 and do not duplicate code.

I think that, like in `Kokkos`, these tests should be disabled for release builds.

common/src/KokkosFFT_utils.hpp
```diff
@@ -8,19 +8,60 @@
 #include <Kokkos_Core.hpp>
 #include <vector>
 #include <set>
+#include <string>
 #include <algorithm>
 #include <numeric>
 #include "KokkosFFT_traits.hpp"
 
+#ifndef KOKKOS_ENABLE_CXX17
```
cedricchevalier19, line 16:
```suggestion
#if defined(KOKKOS_ENABLE_CXX17)
```
Will be easier to extend to other conditions.

cedricchevalier19, line 16: Testing C++20 features with a C++17 macro is not very future-proof.

You can define a `KOKKOSFFT_HAVE_C++20` macro built on top of the Kokkos one.

cedricchevalier19 (commented):
BTW, are you sure the CI tests both C++20 and no C++20?

yasahi-hpc:
> BTW, are you sure the CI tests both C++20 and no C++20?

I was planning to add C++20 after this PR, but it may be better to do that before

common/src/KokkosFFT_utils.hpp
yasahi-hpc: I think it is OK to keep the same logic as `Kokkos`. After `Kokkos` drops C++17, I will just remove C++17, so it should be fine.

common/src/KokkosFFT_utils.hpp
yasahi-hpc: I would like to keep the EXPECTS for `C++17` as well, since the aim is to improve the error messages. 
I am interested in the option 2, but it is not clear to me. Could you please show a snippet for your proposal?

yasahi-hpc: I have fixed accordingly. May I have your second review please

PR #124, opened 2024-07-29
Update base images for gcc and rocm for C++20 build

yasahi-hpc:
This PR aims at updating base images for gcc and rocm in order to build with C++20.

### DIFF
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -12,6 +12,12 @@
 #include <numeric>
 #include "KokkosFFT_traits.hpp"
 
+#if defined(KOKKOS_ENABLE_CXX17)
+#include <cstdlib>
+#else
+#include <source_location>
+#endif
+
 namespace KokkosFFT {
 namespace Impl {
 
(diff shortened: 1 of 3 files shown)


tpadioleau (approved):
Seems good to me. Looking at https://en.cppreference.com/w/cpp/compiler_support, support of `source_location` in libstdc++ has been introduced in GCC 11.

An alternative could be to use feature testing with https://en.cppreference.com/w/cpp/feature_test#cpp_lib_source_location. You could still claim to support C++20 even for older libstdc++ if desired without updating your CI.

yasahi-hpc:
> Seems good to me. Looking at https://en.cppreference.com/w/cpp/compiler_support, support of `source_location` in libstdc++ has been introduced in GCC 11.
> 
> An alternative could be to use feature testing with https://en.cppreference.com/w/cpp/feature_test#cpp_lib_source_location. You could still claim to support C++20 even for older libstdc++ if desired without updating your CI.

Thank you for your comments and approval. 
I would prefer to rely on GCC 11 for simplicity. 

If we use on `feature testing`, I guess the guard macro would be complicated.

PR #125, opened 2024-07-31
Improve check functions

yasahi-hpc:
Improves #80

This PR aims at extending some check functions to work on `std::array` or `std::vector`. A check function `are_valid_axes` has been introduced. For this purpose, a new trait to extract the base value type has been added.

- [x] Add a trait `base_container_value_type` to extract the base value type of a container
- [x] Update existing check functions to work on `std::array` or `std::vector`.
- [x] Introduce a new check function `are_valid_axes` to check if the axes are valid for input Views

### DIFF
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -126,26 +126,27 @@ bool are_valid_axes(const ViewType& view, const ArrayType<IntType, DIM>& axes) {
   static_assert(ViewType::rank() >= DIM,
                 "are_valid_axes: View rank must be larger than or equal to the "
                 "Rank of FFT axes");
+
+  // int type is choosen for consistency with the rest of the code
+  // the axes are defined with int type
   constexpr int rank = ViewType::rank();
 
   // Convert the input axes to be in the range of [0, rank-1]
-  std::vector<int> non_negative_axes;
+  std::array<int, DIM> non_negative_axes;
 
-  // In case of exception, we capture it and return false,
-  // since we do not want to throw an exception here
+  // In case axis is out of range, 'convert_negative_axis' will throw an
+  // runtime_error and we will return false. Without runtime_error, it is
+  // ensured that the 'non_negative_axes' are in the range of [0, rank-1]
   try {
     for (std::size_t i = 0; i < DIM; i++) {
       int axis = KokkosFFT::Impl::convert_negative_axis(view, axes[i]);
-      non_negative_axes.push_back(axis);
+      non_negative_axes[i] = axis;
     }
-  } catch (std::exception& e) {
+  } catch (std::runtime_error& e) {
     return false;
   }
 
-  bool is_valid = (!KokkosFFT::Impl::has_duplicate_values(non_negative_axes)) &&
-                  (!KokkosFFT::Impl::is_out_of_range_value_included(
-                      non_negative_axes, rank));
-
+  bool is_valid = !KokkosFFT::Impl::has_duplicate_values(non_negative_axes);
   return is_valid;
 }
 
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -127,11 +127,9 @@ bool are_valid_axes(const ViewType& view, const ArrayType<IntType, DIM>& axes) {
                 "are_valid_axes: View rank must be larger than or equal to the "
                 "Rank of FFT axes");
 
+  // Convert the input axes to be in the range of [0, rank-1]
   // int type is choosen for consistency with the rest of the code
   // the axes are defined with int type
-  constexpr int rank = ViewType::rank();
-
-  // Convert the input axes to be in the range of [0, rank-1]
   std::array<int, DIM> non_negative_axes;
 
   // In case axis is out of range, 'convert_negative_axis' will throw an
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -130,9 +130,16 @@ bool are_valid_axes(const ViewType& view, const ArrayType<IntType, DIM>& axes) {
 
   // Convert the input axes to be in the range of [0, rank-1]
   std::vector<int> non_negative_axes;
-  for (std::size_t i = 0; i < DIM; i++) {
-    int axis = KokkosFFT::Impl::convert_negative_axis(view, axes[i]);
-    non_negative_axes.push_back(axis);
+
+  // In case of exception, we capture it and return false,
+  // since we do not want to throw an exception here
+  try {
+    for (std::size_t i = 0; i < DIM; i++) {
+      int axis = KokkosFFT::Impl::convert_negative_axis(view, axes[i]);
+      non_negative_axes.push_back(axis);
+    }
+  } catch (std::exception& e) {
+    return false;
   }
 
   bool is_valid = (!KokkosFFT::Impl::has_duplicate_values(non_negative_axes)) &&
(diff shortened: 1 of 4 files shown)


common/src/KokkosFFT_utils.hpp
```diff
@@ -85,38 +86,82 @@ auto convert_negative_shift(const ViewType& view, int _shift, int _axis) {
...
 }
 
-template <typename IntType, std::enable_if_t<std::is_integral_v<IntType>,
-                                             std::nullptr_t> = nullptr>
-bool is_out_of_range_value_included(const std::vector<IntType>& values,
-                                    IntType max) {
+template <
+    typename ContainerType, typename IntType,
+    std::enable_if_t<std::is_integral_v<IntType>, std::nullptr_t> = nullptr>
+bool is_out_of_range_value_included(const ContainerType& values, IntType max) {
```
cedricchevalier19, line 107: Not related but for this function you can do an early exit in the for loop.

common/src/KokkosFFT_utils.hpp
```diff
@@ -85,38 +86,82 @@ auto convert_negative_shift(const ViewType& view, int _shift, int _axis) {
...
+bool are_valid_axes(const ViewType& view, const ArrayType<IntType, DIM>& axes) {
+  static_assert(
+      DIM >= 1 && DIM <= KokkosFFT::MAX_FFT_DIM,
+      "are_valid_axes: the Rank of FFT axes must be between 1 and MAX_FFT_DIM");
+  static_assert(ViewType::rank() >= DIM,
+                "are_valid_axes: View rank must be larger than or equal to the "
+                "Rank of FFT axes");
+  constexpr int rank = ViewType::rank();
+
+  // Convert the input axes to be in the range of [0, rank-1]
```
cedricchevalier19, line 131: Add a comment on why `int` is good datatype choice.

common/src/KokkosFFT_utils.hpp
```diff
@@ -85,38 +86,82 @@ auto convert_negative_shift(const ViewType& view, int _shift, int _axis) {
...
+  std::vector<int> non_negative_axes;
```
cedricchevalier19, line 132: I do not think this `std::vector` is necessary. Its length is known at compile time. You should use an array or something like this if you need "temporary" storage.

common/src/KokkosFFT_utils.hpp
```diff
@@ -85,38 +86,82 @@ auto convert_negative_shift(const ViewType& view, int _shift, int _axis) {
...
+
+  // In case of exception, we capture it and return false,
+  // since we do not want to throw an exception here
+  try {
+    for (std::size_t i = 0; i < DIM; i++) {
+      int axis = KokkosFFT::Impl::convert_negative_axis(view, axes[i]);
+      non_negative_axes.push_back(axis);
```
cedricchevalier19, line 139: You can check for `out of range` in this loop directly.

common/src/KokkosFFT_utils.hpp
```diff
@@ -85,38 +86,82 @@ auto convert_negative_shift(const ViewType& view, int _shift, int _axis) {
...
+    }
+  } catch (std::exception& e) {
+    return false;
+  }
+
+  bool is_valid = (!KokkosFFT::Impl::has_duplicate_values(non_negative_axes)) &&
+                  (!KokkosFFT::Impl::is_out_of_range_value_included(
+                      non_negative_axes, rank));
```
cedricchevalier19, line 147: So the `non_negative_axes` is used only to check for duplication?

common/src/KokkosFFT_utils.hpp
```diff
@@ -85,38 +86,82 @@ auto convert_negative_shift(const ViewType& view, int _shift, int _axis) {
...
+
+  // In case of exception, we capture it and return false,
+  // since we do not want to throw an exception here
+  try {
+    for (std::size_t i = 0; i < DIM; i++) {
+      int axis = KokkosFFT::Impl::convert_negative_axis(view, axes[i]);
+      non_negative_axes.push_back(axis);
+    }
+  } catch (std::exception& e) {
+    return false;
```
cedricchevalier19, line 142: What kind of exception do you expect? It might dangerous to hide all the exceptions.

common/src/KokkosFFT_utils.hpp
yasahi-hpc: That is true. Fixed accordingly

common/src/KokkosFFT_utils.hpp
yasahi-hpc: In the current implementation, axes type is an alias to `std::array<int, DIM>`. Thus, we use `int` here. I have added a comment

common/src/KokkosFFT_utils.hpp
yasahi-hpc: Actually, `out of range` check has already made in `convert_negative_axis` function, which throws a `runtime_error` for `out of range`. Thus, if there is no exception, it is ensured that the axes are not `out of range`. I have added a comment

common/src/KokkosFFT_utils.hpp
yasahi-hpc: As you said, we only need duplication check here. Fixed accordingly.

common/src/KokkosFFT_utils.hpp
yasahi-hpc: We may have `std::runtime_error` here. Fixed to capture `std::runtime_error` here. Also comments to explain the expected behaviors are added

yasahi-hpc:
@cedricchevalier19 Thanks for reviews. I have fixed accordingly. Can I have a second review please

common/src/KokkosFFT_utils.hpp
```diff
@@ -85,43 +86,86 @@ auto convert_negative_shift(const ViewType& view, int _shift, int _axis) {
...
+  std::array<int, DIM> non_negative_axes;
+
+  // In case axis is out of range, 'convert_negative_axis' will throw an
+  // runtime_error and we will return false. Without runtime_error, it is
+  // ensured that the 'non_negative_axes' are in the range of [0, rank-1]
+  try {
+    for (std::size_t i = 0; i < DIM; i++) {
+      int axis = KokkosFFT::Impl::convert_negative_axis(view, axes[i]);
+      non_negative_axes[i] = axis;
+    }
```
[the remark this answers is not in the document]
yasahi-hpc: Yes. We are still storing the non-negative axes which are converted from the original axes. Maybe better to rename this method `convert_negative_axis`, which returns positive `axis` as is and negative axis to be non-negative axis.

Let me explain the definition. `DIM` is dimension of FFT (1D to 3D).
The value `axes` is defined by the `std::array<int, DIM>` to specify the axes to perform FFT. 
Each component of axes can take a range from `[-rank, rank-1] ` where `rank` is a `rank` of `View`.

For example, for 3D and 4D Views we may have the following axes

```C++
axes = {0, 2, -2} // Converted to {0, 2, 1} for 3D View
axes = {0, 2, -1} // Converted to {0, 2, 3} for 4D View
```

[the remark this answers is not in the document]
cedricchevalier19: I do not quite follow the logic here.
```C++
axes = {0, 2, -2}; // {0, 2, 3-2=1} in 3D
axes = {0, 2, 1}; // Same than above?
```
However, I do not think it is related to this PR, I simply have to read the documentation.

cedricchevalier19 (approved):

common/src/KokkosFFT_utils.hpp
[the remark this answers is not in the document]
yasahi-hpc: The logic is the same as `numpy.ftt` but I have not documented this well. I will improve the docs to describe this point.

PR #126, opened 2024-08-20
Improve to print details of tpls in CMake

yasahi-hpc:
Improves #79

This PR aims at improving the CMake logic to include TPLs.

- [x] Fix bugs not to choose correct tpls in case `-DKokkosFFT_ENABLE_HOST_AND_DEVICE=ON` and -`DKokkos_ENABLE_SERIAL=ON` or `DKokkos_ENABLE_THREADS=ON`
- [x] Fix missing #cmakedefine for ROCFFT in KokkosFFT_config.h.in
- [x] Improve messages for the enabled tpls from CMake. It now looks like

```bash
-- KokkosFFT version: 0.1.0
-- KokkosFFT TPLs:
--     Device Parallel: CUFFT
--     Host Parallel: FFTW_OPENMP
--       Host Serial: FFTW_SERIAL
```


cmake/KokkosFFT_tpls.cmake
```diff
@@ -2,34 +2,68 @@
...
+        if(Kokkos_ENABLE_SERIAL AND (${idx_FFTW_SERIAL} EQUAL -1))
+            list(APPEND tpls_list FFTW_SERIAL)
+            list(APPEND backend_list "  Host Serial")
+        endif()
+    endif()
+
+    # Enable Serial if Device is not used
+    list(FIND tpls_list "FFTW_SERIAL" idx_FFTW_SERIAL)
+    if(NOT ${KOKKOSFFT_HAS_DEVICE})
+        if(Kokkos_ENABLE_SERIAL AND (${idx_FFTW_SERIAL} EQUAL -1))
```
pzehner, line 60: Same as above. Also, collapse the two `if`s.

cmake/KokkosFFT_tpls.cmake
```diff
@@ -2,34 +2,68 @@
 #
 # SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
 
-function(get_tpls_list tpls_list)
+function(get_tpls_list tpls_list backend_list)
```
pzehner, line 5: Please improve the documentation of the function to tell what it does, and what the output variables are about. There are no official documentation formats for CMake functions, but [this unofficial one](http://doc.aldebaran.com/1-14/qibuild/contrib/cmake/coding_guide.html) was pretty useful.

yasahi-hpc:
> This is nice. My main concerns are about not using a more recent, clearer CMake syntax, that makes the code easier to understand and to maintain.
> 
> More comments in the code would be appreciated to tell what you want to achieve.
> 
Thank you for the reviews. I will fix accordingly. 

> As for the feature, you added the backend type when displaying the FFT library to use, and made another list to store it. This list is only used for display, so I wonder if it could be deduced from the first one, when displaying it.

For me, iterating over lists is the easiest solution (even better with Zip iterators). Do you have a better solution in mind?

cmake/KokkosFFT_tpls.cmake
yasahi-hpc: Not sure what is the right way. But I added a cstyle docstring

pzehner (approved):
Sounds good to me.

yasahi-hpc:
Good. I will merge this

PR #129, opened 2024-08-29
Apply check functions to common functions

yasahi-hpc:
Improves #80 

This PR aims at replacing runtime `assert` with `KOKKOSFFT_EXPECTS`. Adding more `static_assertions` if applicable.
Modifications are applied to source code under `common/src` and `common/unit_test`.

- [x] Apply `are_valid_axes` function where it is applicable
- [x] Real to real case is suppressed by `static_assertion` in `KokkosFFT_layout.hpp`
- [x] Apply `KOKKOSFFT_EXPECTS` to some if else conditional (throw `std::runtime_error` in the case of `else`)
- [x] Replacing `is_view<ViewType>::value` with `is_view_v<ViewType>`

### DIFF
--- a/common/src/KokkosFFT_normalization.hpp
+++ b/common/src/KokkosFFT_normalization.hpp
@@ -62,6 +62,11 @@ template <typename ExecutionSpace, typename ViewType>
 void normalize(const ExecutionSpace& exec_space, ViewType& inout,
                Direction direction, Normalization normalization,
                std::size_t fft_size) {
+  static_assert(KokkosFFT::Impl::is_operatable_view_v<ExecutionSpace, ViewType>,
+                "normalize: View value type must be float, double, "
+                "Kokkos::Complex<float>, or Kokkos::Complex<double>. "
+                "Layout must be either LayoutLeft or LayoutRight. "
+                "ExecutionSpace must be able to access data in ViewType");
   auto [coef, to_normalize] =
       get_coefficients(inout, direction, normalization, fft_size);
   if (to_normalize) normalize_impl(exec_space, inout, coef);
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -54,7 +54,7 @@ inline void check_precondition(const bool expression,
 
 template <typename ViewType>
 auto convert_negative_axis(ViewType, int _axis = -1) {
-  static_assert(Kokkos::is_view<ViewType>::value,
+  static_assert(Kokkos::is_view_v<ViewType>,
                 "convert_negative_axis: ViewType is not a Kokkos::View.");
   int rank = static_cast<int>(ViewType::rank());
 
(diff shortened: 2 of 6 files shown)


common/src/KokkosFFT_normalization.hpp
```diff
@@ -62,6 +62,11 @@ template <typename ExecutionSpace, typename ViewType>
 void normalize(const ExecutionSpace& exec_space, ViewType& inout,
                Direction direction, Normalization normalization,
                std::size_t fft_size) {
+  static_assert(KokkosFFT::Impl::is_operatable_view_v<ExecutionSpace, ViewType>,
+                "normalize: View value type must be float, double, "
+                "Kokkos::Complex<float>, or Kokkos::Complex<double>. "
+                "Layout must be either LayoutLeft or LayoutRight. "
+                "ExecutionSpace must be able to access data in ViewType");
```
tpadioleau, line 69: Is it better to merge all the checks in one `static_assert` or split them in multiple ones ?

common/src/KokkosFFT_utils.hpp
```diff
@@ -118,14 +118,12 @@ bool is_out_of_range_value_included(const ContainerType& values, IntType max) {
 template <
     typename ViewType, template <typename, std::size_t> class ArrayType,
     typename IntType, std::size_t DIM = 1,
-    std::enable_if_t<std::is_integral_v<IntType>, std::nullptr_t> = nullptr>
+    std::enable_if_t<Kokkos::is_view_v<ViewType> && std::is_integral_v<IntType>,
+                     std::nullptr_t> = nullptr>
```
tpadioleau, line 122: Do you have other overloads for this function ? If not I would simply use a static assertion instead of SFINAE.

tpadioleau (changes requested):
Seems globally good to me!

Not directly related to his PR but I see you rely on the standard version to use `std::source_location`. I think it is more reliable to use feature testing like `#if defined(__cpp_lib_source_location)`. This way you would be compatible with a compiler that supports only a part of C++ 20.

common/src/KokkosFFT_utils.hpp
yasahi-hpc: True. I have fixed accordingly.

common/src/KokkosFFT_normalization.hpp
yasahi-hpc: That may be changed in the future. For the moment, I would like to merge all checks in order not to forget some of them.

yasahi-hpc:
> Seems globally good to me!

Thank you for your reviews!

> 
> Not directly related to his PR but I see you rely on the standard version to use `std::source_location`. I think it is more reliable to use feature testing like `#if defined(__cpp_lib_source_location)`. This way you would be compatible with a compiler that supports only a part of C++ 20.

I have fixed accordingly. My aims was to make sure that C++ 20 build actually relies on `std::source_location` rather than using C++17 version in case compiler does not support it.

PR #130, opened 2024-09-06
Apply check functions to fft functions

yasahi-hpc:
Improves #80 

This PR aims at replacing runtime `assert` with `KOKKOSFFT_EXPECTS`. Adding more `static_assertions` if applicable.
Modifications are applied to source code under `fft/src` and `fft/unit_test`.

- [x] Move `KOKKOSFFT_EXPECTS` from `KokkosFFT_utils.hpp` to `KokkosFFT_asserts.hpp`
- [x] Apply `are_valid_axes` function where it is applicable
- [x] Apply `KOKKOSFFT_EXPECTS` to some if else conditional (throw `std::runtime_error` in the case of `else`)
- [x] Replacing `is_complex<ViewType>::value` with `is_complex_v<ViewType>`

### DIFF
--- a/common/src/KokkosFFT_asserts.hpp
+++ b/common/src/KokkosFFT_asserts.hpp
@@ -1,0 +1,54 @@
+// SPDX-FileCopyrightText: (C) The Kokkos-FFT development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+#ifndef KOKKOSFFT_ASSERTS_HPP
+#define KOKKOSFFT_ASSERTS_HPP
+
+#include <stdexcept>
+#include <sstream>
+#include <string>
+
+#if defined(__cpp_lib_source_location) && __cpp_lib_source_location >= 201907L
+#include <source_location>
+#define KOKKOSFFT_EXPECTS(expression, msg)                            \
+  KokkosFFT::Impl::check_precondition(                                \
+      (expression), msg, std::source_location::current().file_name(), \
+      std::source_location::current().line(),                         \
+      std::source_location::current().function_name(),                \
+      std::source_location::current().column())
+#else
+#include <cstdlib>
+#define KOKKOSFFT_EXPECTS(expression, msg)                                   \
+  KokkosFFT::Impl::check_precondition((expression), msg, __FILE__, __LINE__, \
+                                      __FUNCTION__)
+#endif
+
+namespace KokkosFFT {
+namespace Impl {
+
+inline void check_precondition(const bool expression,
+                               [[maybe_unused]] const std::string& msg,
+                               [[maybe_unused]] const char* file_name, int line,
+                               [[maybe_unused]] const char* function_name,
+                               [[maybe_unused]] const int column = -1) {
+  // Quick return if possible
+  if (expression) return;
+
+  std::stringstream ss("file: ");
+  if (column == -1) {
+    // For C++ 17
+    ss << file_name << '(' << line << ") `" << function_name << "`: " << msg
+       << '\n';
+  } else {
+    // For C++ 20 and later
+    ss << file_name << '(' << line << ':' << column << ") `" << function_name
+       << "`: " << msg << '\n';
+  }
+  throw std::runtime_error(ss.str());
+}
+
+}  // namespace Impl
+}  // namespace KokkosFFT
+
+#endif
--- a/common/src/KokkosFFT_asserts.hpp
+++ b/common/src/KokkosFFT_asserts.hpp
@@ -7,7 +7,7 @@
 
 #include <stdexcept>
 #include <sstream>
-#include <string>
+#include <string_view>
 
 #if defined(__cpp_lib_source_location) && __cpp_lib_source_location >= 201907L
 #include <source_location>
@@ -28,10 +28,10 @@ namespace KokkosFFT {
 namespace Impl {
 
 inline void check_precondition(const bool expression,
-                               [[maybe_unused]] const std::string& msg,
-                               [[maybe_unused]] const char* file_name, int line,
-                               [[maybe_unused]] const char* function_name,
-                               [[maybe_unused]] const int column = -1) {
+                               const std::string_view& msg,
+                               const char* file_name, int line,
+                               const char* function_name,
+                               const int column = -1) {
   // Quick return if possible
   if (expression) return;
 
--- a/fft/src/KokkosFFT_Plans.hpp
+++ b/fft/src/KokkosFFT_Plans.hpp
@@ -172,18 +172,19 @@ class Plan {
     static_assert(InViewType::rank() >= 1,
                   "Plan::Plan: View rank must be larger than or equal to 1");
 
-    if (KokkosFFT::Impl::is_real_v<in_value_type> &&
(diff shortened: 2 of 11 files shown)


common/src/KokkosFFT_asserts.hpp
```diff
@@ -0,0 +1,54 @@
...
+#ifndef KOKKOSFFT_ASSERTS_HPP
+#define KOKKOSFFT_ASSERTS_HPP
+
+#include <stdexcept>
+#include <sstream>
+#include <string>
+
+#if defined(__cpp_lib_source_location) && __cpp_lib_source_location >= 201907L
+#include <source_location>
+#define KOKKOSFFT_EXPECTS(expression, msg)                            \
```
tpadioleau, line 14: Not a modification from this PR.

I guess the naming is inspired from Kokkos ? In Kokkos, https://github.com/kokkos/kokkos/blob/08ceff92bcf3a828844480bc1e6137eb74028517/core/src/Kokkos_Assert.hpp#L59, these checks are disabled in non-debug, shall we also do that ?

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -233,18 +234,18 @@ class Plan {
...
 
-    if (KokkosFFT::Impl::is_real_v<out_value_type> &&
-        m_direction != KokkosFFT::Direction::backward) {
-      throw std::runtime_error(
-          "Plan::Plan: complex to real transform is constrcuted with forward "
-          "direction.");
+    if constexpr (KokkosFFT::Impl::is_real_v<out_value_type>) {
+      KOKKOSFFT_EXPECTS(
+          m_direction == KokkosFFT::Direction::backward,
+          "complex to real transform is constrcuted with forward direction.");
```
tpadioleau, line 248:
```suggestion
          "complex to real transform is constructed with forward direction.");
```

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -233,18 +234,18 @@ class Plan {
...
-        m_direction != KokkosFFT::Direction::forward) {
-      throw std::runtime_error(
-          "Plan::Plan: real to complex transform is constrcuted with backward "
-          "direction.");
+    KOKKOSFFT_EXPECTS(KokkosFFT::Impl::are_valid_axes(in, m_axes),
+                      "axes are invalid for in/out views");
+    if constexpr (KokkosFFT::Impl::is_real_v<in_value_type>) {
+      KOKKOSFFT_EXPECTS(
+          m_direction == KokkosFFT::Direction::forward,
+          "real to complex transform is constrcuted with backward direction.");
```
tpadioleau, line 242:
```suggestion
          "real to complex transform is constructed with backward direction.");
```

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -172,18 +172,19 @@ class Plan {
...
-      throw std::runtime_error(
-          "Plan::Plan: real to complex transform is constrcuted with backward "
-          "direction.");
+    KOKKOSFFT_EXPECTS(KokkosFFT::Impl::are_valid_axes(in, m_axes),
+                      "axes are invalid for in/out views");
+
+    if constexpr (KokkosFFT::Impl::is_real_v<in_value_type>) {
+      KOKKOSFFT_EXPECTS(
+          m_direction == KokkosFFT::Direction::forward,
+          "real to complex transform is constrcuted with backward direction.");
```
tpadioleau, line 181:
```suggestion
          "real to complex transform is constructed with backward direction.");
```

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -172,18 +172,19 @@ class Plan {
...
 
-    if (KokkosFFT::Impl::is_real_v<out_value_type> &&
-        m_direction != KokkosFFT::Direction::backward) {
-      throw std::runtime_error(
-          "Plan::Plan: complex to real transform is constrcuted with forward "
-          "direction.");
+    if constexpr (KokkosFFT::Impl::is_real_v<out_value_type>) {
+      KOKKOSFFT_EXPECTS(
+          m_direction == KokkosFFT::Direction::backward,
+          "complex to real transform is constrcuted with forward direction.");
```
tpadioleau, line 187:
```suggestion
          "complex to real transform is constructed with forward direction.");
```

tpadioleau (changes requested):
My main question is about whether to enable the checks in non-debug mode or not ?

fft/src/KokkosFFT_Plans.hpp
yasahi-hpc: Fixed accordingly

yasahi-hpc:
> My main question is about whether to enable the checks in non-debug mode or not ?

First of all, thank you for your reviews. Actually, we need it to enable for non-debug mode (different from Kokkos). The point is that if we execute an internal fft library without satisfying the conditions, we may get unexpected errors from the library. I would like to let users know that input data are invalid with the appropriate error messages.

common/src/KokkosFFT_asserts.hpp
```diff
@@ -0,0 +1,54 @@
...
+#define KOKKOSFFT_EXPECTS(expression, msg)                                   \
+  KokkosFFT::Impl::check_precondition((expression), msg, __FILE__, __LINE__, \
+                                      __FUNCTION__)
+#endif
+
+namespace KokkosFFT {
+namespace Impl {
+
+inline void check_precondition(const bool expression,
+                               [[maybe_unused]] const std::string& msg,
```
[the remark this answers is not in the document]
yasahi-hpc: It would be better. I have changed accordingly.

common/src/KokkosFFT_asserts.hpp
yasahi-hpc: Maybe change the name in the future. This check is stricter than Kokkos one in order to avoid random errors from invalid usage of fft libraries.

tpadioleau: I understand the purpose and agree with it. What do you think of `KOKKOSFFT_THROW_IF` ?

yasahi-hpc: That makes sense. 
So, `KOKKOSFFT_THROW_IF` throws if `expression` is satisfied?
Then we need to invert conditionals.

tpadioleau: > So, `KOKKOSFFT_THROW_IF` throws if `expression` is satisfied? Then we need to invert conditionals.

Yes that would be it.

A second reason to rename is that Kokkos aborts whereas you throw an exception. You let the user a chance to catch the error whereas Kokkos does not.

yasahi-hpc: Sure. 
The reason why we do not abort is that many FFT libraries do not abort but throw an exception (or return an error state).

tpadioleau: Seems ok for me

yasahi-hpc: I would like to make the change from `KOKKOSFFT_EXPECTS` to `KOKKOSFFT_THROW_IF` in another PR, since it is big and I may make a mistake. Is it fine with you?

tpadioleau: Sure!

PR #135, opened 2024-09-23
Do not install CMakeLists.txt

### LINKED ISSUE #134: Suspicious installed CMakeLists.txt file
In the installed directory `include` I noticed there is a `CMakeLists.txt` file that is installed. I don't think this file should be installed.

yasahi-hpc:
Fixes #134 

In the installed directory `include`, `CMakeLists.txt` should not be installed


tpadioleau (approved):
Thanks!

PR #137, opened 2024-09-24
fix: cmake install with --prefix option

### LINKED ISSUE #136: Cannot install when specifying `--prefix` on the command line
When installing with `cmake --install build --prefix some/prefix/path`, it still tries to install in `CMAKE_INSTALL_PREFIX` which defaults to `/usr/local` on my system. It seems to work fine when specifying `CMAKE_INSTALL_PREFIX`.

To reproduce:
```cmake
cmake -DCMAKE_INSTALL_PREFIX=$PWD/install-cmakevar-prefix -B build
cmake --build build
cmake --install build --prefix $PWD/install-cmdline-prefix
```
The directory `$PWD/install-cmdline-prefix` is empty while it should not.

I suspect it comes from variables that read `CMAKE_INSTALL_PREFIX` at configuration time and store the result like
https://github.com/kokkos/kokkos-fft/blob/49200b82ceaf9125b59e23b5fb0867c8a307edda/CMakeLists.txt#L117
These are not updated when `--prefix` is used.

yasahi-hpc:
Fixes #136 

Use paths relative to `${CMAKE_INSTALL_PREFIX}`


yasahi-hpc:
> It works for me, thanks!

That is nice!
Thank you for confirmation. I will merge this

PR #138, opened 2024-09-24
replace std::copy with std::copy_n to avoid -Warray-bounds issue

yasahi-hpc:
This PR aims at suppressing unexpected `-Warray-bounds issue` error from `std::copy` between `std::vector` and `std::array`. `std::copy_n` seems to suppress this issue.

- [x] Replace `std::copy` with `std::copy_n`
- [x] `_axes` and `axes` are renamed to `axes` and `non_negative_axes`
- [x] `non_negative_axes` relies on `std::array` rather than `std::vector`.

### DIFF
--- a/common/src/KokkosFFT_transpose.hpp
+++ b/common/src/KokkosFFT_transpose.hpp
@@ -20,8 +20,8 @@ auto get_map_axes(const ViewType& view, axis_type<DIM> axes) {
   // Convert the input axes to be in the range of [0, rank-1]
   axis_type<DIM> non_negative_axes = {};
   for (std::size_t i = 0; i < DIM; i++) {
-    int axis = KokkosFFT::Impl::convert_negative_axis(view, axes.at(i));
-    non_negative_axes[i] = axis;
+    non_negative_axes.at(i) =
+        KokkosFFT::Impl::convert_negative_axis(view, axes.at(i));
   }
 
   // how indices are map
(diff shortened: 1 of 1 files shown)


common/src/KokkosFFT_transpose.hpp
```diff
@@ -13,62 +13,60 @@
...
                      "get_map_axes: input axes are not valid for the view");
 
   // Convert the input axes to be in the range of [0, rank-1]
-  std::vector<int> axes;
+  axis_type<DIM> non_negative_axes = {};
   for (std::size_t i = 0; i < DIM; i++) {
-    int axis = KokkosFFT::Impl::convert_negative_axis(view, _axes.at(i));
-    axes.push_back(axis);
+    int axis = KokkosFFT::Impl::convert_negative_axis(view, axes.at(i));
+    non_negative_axes[i] = axis;
```
tpadioleau, line 24:
```suggestion
    non_negative_axes[i] = KokkosFFT::Impl::convert_negative_axis(view, axes.at(i));
```

Seems ok to do it at once ?

tpadioleau (approved):
Ok for me

common/src/KokkosFFT_transpose.hpp
yasahi-hpc: Certainly, I will do that

tpadioleau (approved):
Ok for me

PR #139, opened 2024-09-24
Remove unused lines

yasahi-hpc:
This PR aims at cleaning up. Following modifications are made.

- [x] Using new static assertions for operation
- [x] Remove an unused argument from `fft_exec_impl` function


tpadioleau (approved):
Ok for me, one question though, the string in the `static_assert` is quite long, do you think we could factorize this `static_assert` call into an empty function ?

yasahi-hpc:
> Ok for me, one question though, the string in the `static_assert` is quite long, do you think we could factorize this `static_assert` call into an empty function ?

Actually, I am planning to optimize the error messages in #133

PR #140, opened 2024-09-24
Allow user to work on const output View

yasahi-hpc:
This PR aims at allowing users to work on const output View type.

For example,

```C++
#include <KokkosFFT.hpp>
#include <Kokkos_Core.hpp>

int main() {
  Kokkos::ScopeGuard scope;
  Kokkos::View<Kokkos::complex<double> **> const in_view("in", 10, 10);
  Kokkos::View<Kokkos::complex<double> **> const out_view("out", 10, 10);
  KokkosFFT::fft2(Kokkos::DefaultExecutionSpace(), in_view, out_view);
  return 0;
}
```

### DIFF
--- a/common/src/KokkosFFT_transpose.hpp
+++ b/common/src/KokkosFFT_transpose.hpp
@@ -91,9 +91,10 @@ void prep_transpose_view(InViewType& in, OutViewType& out,
   }
 
   if constexpr (std::is_const_v<OutViewType>) {
-    KOKKOSFFT_THROW_IF(!is_out_view_ready,
-                       "prep_transpose_view: OutViewType is const, but do not "
-                       "have the required extents");
+    KOKKOSFFT_THROW_IF(
+        !is_out_view_ready,
+        "prep_transpose_view: OutViewType is const, but does not "
+        "have the required extents");
   } else {
     if (!is_out_view_ready) {
       if constexpr (!OutViewType::memory_traits::is_unmanaged) {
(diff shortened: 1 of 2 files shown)


common/src/KokkosFFT_transpose.hpp
```diff
@@ -90,12 +90,18 @@ void prep_transpose_view(InViewType& in, OutViewType& out,
...
 
-  if (!is_out_view_ready) {
-    if constexpr (!OutViewType::memory_traits::is_unmanaged) {
-      KokkosFFT::Impl::create_view(out, "out", out_extents);
-    } else {
-      // try to reshape out if it currently has enough memory available
-      KokkosFFT::Impl::reshape_view(out, out_extents);
+  if constexpr (std::is_const_v<OutViewType>) {
+    KOKKOSFFT_THROW_IF(!is_out_view_ready,
+                       "prep_transpose_view: OutViewType is const, but do not "
```
masterleinad, line 95:
```suggestion
                       "prep_transpose_view: OutViewType is const, but does not "
```

tpadioleau (dismissed):
Seems ok to me!

common/src/KokkosFFT_transpose.hpp
yasahi-hpc: Thank you! Fixed accordingly

tpadioleau (approved):

PR #141, opened 2024-09-24
Set default axes for N-dimensional fft APIs

yasahi-hpc:
In this PR, we set default axes for ND fft APIs. 
For example, for 2D Views, default `axes = {-2, -1}`.

```C++
using axes_type = KokkosFFT::axis_type<2>;
axes_type axes  = {-2, -1};
KokkosFFT::fftn(execution_space(), x, out);
KokkosFFT::fftn(execution_space(), x, out, axes);
```

PR #143, opened 2024-10-01
Fix unresolved doxygen for N-dimensional FFT APIs

yasahi-hpc:
This PR aims at fixing docstrings bugs from #141 
As a temporary solution, we introduce a macro for doxygen which can be resolved. See https://github.com/kokkos/kokkos-kernels/pull/1319 for detail.

### DIFF
--- a/fft/src/KokkosFFT_Transform.hpp
+++ b/fft/src/KokkosFFT_Transform.hpp
@@ -469,12 +469,12 @@ void irfft2(const ExecutionSpace& exec_space, const InViewType& in,
 /// \param exec_space [in] Kokkos execution space
 /// \param in [in] Input data (complex)
 /// \param out [out] Ouput data (complex)
-/// \param axes [in] Axes over which FFT is performed (optional)
+/// \param axes [in] Axes over which FFT is performed (default, all axes)
 /// \param norm [in] How the normalization is applied (optional)
 /// \param s [in] Shape of the transformed axis of the output (optional)
 #if defined(DOXY)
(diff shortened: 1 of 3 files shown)


fft/src/KokkosFFT_Transform.hpp
```diff
@@ -469,11 +469,16 @@ void irfft2(const ExecutionSpace& exec_space, const InViewType& in,
 /// \param exec_space [in] Kokkos execution space
 /// \param in [in] Input data (complex)
 /// \param out [out] Ouput data (complex)
-/// \param axes [in] Axes over which FFT is performed (optional)
+/// \param axes [in] Axes over which FFT is performed (default, all axes)
```
tpadioleau, line 472: I think the word `default` is better than `optional`. I mean one cannot only specify the parameter `s` without specifying `axes` and `norm`. The full set `axes`,`norm` and `s` can be interpreted as optional but not really separately ?

tpadioleau (approved):
LGTM

fft/src/KokkosFFT_Transform.hpp
tpadioleau: I forgot a concluding remark, should we consider updating all `(optional)` to `(default, ...)` ?

yasahi-hpc:
@tpadioleau Thank you for your review.

fft/src/KokkosFFT_Transform.hpp
tpadioleau: sure

PR #101, opened 2024-05-23
Implemented passing unmanaged views

helloworld922:
For the output reverse transpose this will try to reshape a given unmanaged buffer if it fits, otherwise throws an exception.

PR #144, opened 2024-10-01
Update docs

yasahi-hpc:
This PR aims at improving the documentation.
Following changes are made.

- [x] Add explanations for default arguments rather than just saying it is optional
- [x] Add a link to unmanaged view example introduced in #101
- [x] Claim that runtime errors are all given with C++ ``std::runtime_error`` (not assertion)
- [x] Add explanation about the meanings of axes which may be non-trivial for non-python users


docs/intro/using.rst
```diff
@@ -130,3 +130,43 @@ In some backend, FFT plan creation leads to some overhead, wherein we need this
...
+   template <typename T> using View2D = Kokkos::View<T**, Kokkos::LayoutLeft, execution_space>;
+   template <typename T> using View3D = Kokkos::View<T***, Kokkos::LayoutLeft, execution_space>;
+   constexpr int n0 = 4, n1 = 8, n2 = 5;
+
+   View2D<double> x2("x2", n0, n1);
+   View3D<double> x3("x3", n0, n1, n2);
+   View2D<Kokkos::complex<double> > x2_hat("x2_hat", n0/2+1, n1);
+   View3D<Kokkos::complex<double> > x3_hat("x3_hat", n0, n1/2+1, n2);
+
+   // Follwoing codes are all equivalent to np.fft(np.rfft(axis=0), axis=1)
```
tpadioleau, line 157:
```suggestion
   // Following codes are all equivalent to np.fft(np.rfft(axis=0), axis=1)
```

tpadioleau (approved):
LGTM

fft/src/KokkosFFT_Transform.hpp
```diff
@@ -123,9 +123,10 @@ namespace KokkosFFT {
 /// \param exec_space [in] Kokkos execution space
 /// \param in [in] Input data (complex)
 /// \param out [out] Ouput data (complex)
-/// \param norm [in] How the normalization is applied (optional)
-/// \param axis [in] Axis over which FFT is performed (optional)
-/// \param n [in] Length of the transformed axis of the output (optional)
+/// \param norm [in] How the normalization is applied (default, backward)
+/// \param axis [in] Axis over which FFT is performed (default, -1)
```
tpadioleau, line 127: Actually I have one question, why do we make a difference between `norm`/`axis` and `n` ? Only the last parameter is a `std::optional`.

yasahi-hpc: Basically, two reasons.
1. Firstly, this parameter may not be frequently used. So, I would like to set `None` as a default parameter.
2. Secondly, we cannot give a default value in compile time (cannot hard-code), which is affected by the in/out view extents. In contrast, `norm` and `axis` can have fixed values.

tpadioleau: > norm and axis can have fixed values.

We could uniformly use `std::optional` ? In 2D for example,

```cpp
template <typename ExecutionSpace, typename InViewType, typename OutViewType>
void fft2(const ExecutionSpace& exec_space, const InViewType& in,
          OutViewType& out,
          std::optional<KokkosFFT::Normalization> norm = std::nullopt,
          std::optional<axis_type<2>> axes = std::nullopt, std::optional<shape_type<2>> s = std::nullopt)
```

This would allow to specify only one parameter such as

```cpp
fft(exec_space, in, out, std:::nullopt, {1, 0})
```

yasahi-hpc: We can also do that.
The current implementation helps doxygen to auto generate default parameters, for example,

```C++
template<typename ExecutionSpace, typename InViewType, typename OutViewType>
void KokkosFFT::fft(
  const ExecutionSpace &exec_space,
  const &in,
  OutViewType &out,
  KokkosFFT::Normalization norm = KokkosFFT::Normalization::backward,
  int axis = -1,
  std::optional<std::size_t> n = std::nullopt)
```

tpadioleau: Closing, this discussion is not critical for this PR.

common/src/KokkosFFT_Helpers.hpp
```diff
@@ -215,7 +215,7 @@ auto rfftfreq(const ExecutionSpace&, const std::size_t n,
 ///
 /// \param exec_space [in] Kokkos execution space
 /// \param inout [in,out] Spectrum
-/// \param axes [in] Axes over which to shift, optional
+/// \param axes [in] Axes over which to shift (default, nullopt)
```
cedricchevalier19, line 218: Can you explain or point out the default behavior with `nullopt`?

common/src/KokkosFFT_Helpers.hpp
```diff
@@ -264,7 +264,7 @@ void fftshift(const ExecutionSpace& exec_space, ViewType& inout,
...
```
cedricchevalier19, line 267: Same here

docs/intro/using.rst
```diff
@@ -130,3 +130,43 @@ In some backend, FFT plan creation leads to some overhead, wherein we need this
...
+   View2D<double> x2("x2", n0, n1);
+   View3D<double> x3("x3", n0, n1, n2);
+   View2D<Kokkos::complex<double> > x2_hat("x2_hat", n0/2+1, n1);
+   View3D<Kokkos::complex<double> > x3_hat("x3_hat", n0, n1/2+1, n2);
+
+   // Following codes are all equivalent to np.fft(np.rfft(axis=0), axis=1)
+   KokkosFFT::rfft2(execution_space(), x2, x2_hat, /*axes=*/{-1, -2});
+   KokkosFFT::rfft2(execution_space(), x2, x2_hat, /*axes=*/{-1, 0});
+   KokkosFFT::rfft2(execution_space(), x2, x2_hat, /*axes=*/{1, -2});
+   KokkosFFT::rfft2(execution_space(), x2, x2_hat, /*axes=*/{1, 0});
```
cedricchevalier19, line 161: Perhaps add comment on how negative numbers are converted like `Rank(2)-2 = 0`, `Rank(2)-1 = 1`

cedricchevalier19 (commented):
Looks better.

docs/intro/using.rst
yasahi-hpc: That is a good point. I added accordingly.

docs/intro/using.rst
```diff
@@ -130,3 +130,48 @@ In some backend, FFT plan creation leads to some overhead, wherein we need this
...
+   // negative axes are converted as follows:
+   // -2 -> 1 (= Rank(3) - 2), -1 -> 2 (= Rank(3) - 1)
+   KokkosFFT::rfft2(execution_space(), x3, x3_hat, /*axes=*/{-1, -2});
+   KokkosFFT::rfft2(execution_space(), x3, x3_hat, /*axes=*/{-1, 1});
+   KokkosFFT::rfft2(execution_space(), x3, x3_hat, /*axes=*/{2, -2});
+   KokkosFFT::rfft2(execution_space(), x3, x3_hat, /*axes=*/{2, 1});
+
+.. note::
+
+   If you rely on negative axes, the corresponding positve axes are different depending on the rank of Views.
```
acalloo, line 176: positive

yasahi-hpc: Good catch. I also modified the sentence a little bit

acalloo (approved):

PR #145, opened 2024-10-02
Remove duplicated C++17 flags

tpadioleau:
I think these flags are redundant with the compile features above


yasahi-hpc:
Thank you for this PR. I do not remember why I put this

yasahi-hpc (approved):

PR #155, opened 2024-10-08
Rename complex number to avoid confusion with the imaginary unit

tpadioleau:
I suspect a typo about the value of `I` which I guess represent the imaginary unit ?

### DIFF
--- a/examples/01_1DFFT/01_1DFFT.cpp
+++ b/examples/01_1DFFT/01_1DFFT.cpp
@@ -15,7 +15,7 @@ int main(int argc, char* argv[]) {
   Kokkos::initialize(argc, argv);
   {
     constexpr int n0 = 128;
-    const Kokkos::complex<double> I(1.0, 1.0);
+    const Kokkos::complex<double> z(1.0, 1.0);
 
     // 1D C2C FFT (Forward and Backward)
     View1D<Kokkos::complex<double> > xc2c("xc2c", n0);
@@ -24,7 +24,7 @@ int main(int argc, char* argv[]) {
 
     Kokkos::Random_XorShift64_Pool<> random_pool(12345);
     execution_space exec;
-    Kokkos::fill_random(exec, xc2c, random_pool, I);
+    Kokkos::fill_random(exec, xc2c, random_pool, z);
     exec.fence();
 
     KokkosFFT::fft(exec, xc2c, xc2c_hat);
@@ -43,7 +43,7 @@ int main(int argc, char* argv[]) {
     // 1D C2R FFT
     View1D<Kokkos::complex<double> > xc2r("xr2c_hat", n0 / 2 + 1);
     View1D<double> xc2r_hat("xc2r", n0);
-    Kokkos::fill_random(exec, xc2r, random_pool, I);
+    Kokkos::fill_random(exec, xc2r, random_pool, z);
     exec.fence();
 
     KokkosFFT::irfft(exec, xc2r, xc2r_hat);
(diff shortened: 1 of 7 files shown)


yasahi-hpc:
Actually, this value is nothing to do with the imaginary unit. So it should be renamed. 
This value `Kokkos::complex<double>(1.0, 1.0)` is intended to get a random complex random number `x` which satisfies`Kokkos::complex<double>(0.0, 0.0) < x < Kokkos::complex<double>(1.0, 1.0)`.

tpadioleau:
> Actually, this value is nothing to do with the imaginary unit. So it should be renamed.

I see, what do you think of `z` or `c`, I used to see a lot this naming in math lessons ?

> This value `Kokkos::complex<double>(1.0, 1.0)` is intended to get a random complex random number `x` which satisfies`Kokkos::complex<double>(0.0, 0.0) < x < Kokkos::complex<double>(1.0, 1.0)`.

You mean a random number on the segment $\\{t (1 + i), t \in [0, 1]\\}$ ? Actually I didn't find any information about `fill_random`.

yasahi-hpc:
> > Actually, this value is nothing to do with the imaginary unit. So it should be renamed.
> 
> I see, what do you think of `z` or `c`, I used to see a lot this naming in math lessons ?

I personally prefer `z`, but nothing is wrong with `c`.
 
> > This value `Kokkos::complex<double>(1.0, 1.0)` is intended to get a random complex random number `x` which satisfies`Kokkos::complex<double>(0.0, 0.0) < x < Kokkos::complex<double>(1.0, 1.0)`.
> 
> You mean a random number on the segment { t ( 1 + i ) , t ∈ [ 0 , 1 ] } ? Actually I didn't find any information about `fill_random`.

Exactly. According to the docstring in the [source code](https://github.com/kokkos/kokkos/blob/4e88587265240078b2d6d94261d118e7a4a15cea/algorithms/src/Kokkos_Random.hpp#L200-L202), it would return the random numbers of the type in the range `[0,range)`

tpadioleau:
> I personally prefer `z`, but nothing is wrong with `c`.

I will rename accordingly.

> Exactly. According to the docstring in the [source code](https://github.com/kokkos/kokkos/blob/4e88587265240078b2d6d94261d118e7a4a15cea/algorithms/src/Kokkos_Random.hpp#L200-L202), it would return the random numbers of the type in the range `[0,range)`

I find the notation `[0, range)` a bit ambiguous for complex numbers as there is no natural order for complex numbers.

tpadioleau:
done 4693fada061a6101d6ec08e4c34d81e58f190e18

yasahi-hpc:
> > I personally prefer `z`, but nothing is wrong with `c`.
> 
> I will rename accordingly.

Thanks!

> > Exactly. According to the docstring in the [source code](https://github.com/kokkos/kokkos/blob/4e88587265240078b2d6d94261d118e7a4a15cea/algorithms/src/Kokkos_Random.hpp#L200-L202), it would return the random numbers of the type in the range `[0,range)`
> 
> I find the notation `[0, range)` a bit ambiguous for complex numbers as there is no natural order for complex numbers.

That is true. Actually, I could not find the detailed documentations for this function.

yasahi-hpc (approved):
LGTM. 
I will merge this

tpadioleau:
> Exactly. According to the docstring in the [source code](https://github.com/kokkos/kokkos/blob/4e88587265240078b2d6d94261d118e7a4a15cea/algorithms/src/Kokkos_Random.hpp#L200-L202), it would return the random numbers of the type in the range `[0,range)`

Actually it seems to generate random numbers for each component, real and imaginary, see https://github.com/kokkos/kokkos/blob/4e88587265240078b2d6d94261d118e7a4a15cea/algorithms/src/Kokkos_Random.hpp#L544-L551

Issue #158, opened 2024-10-08
Consider passing output views by `const&`


tpadioleau:
I don't think we want our API to handle allocations of output views ? I would suggest to pass output views by `const&` so that the user is guaranteed KokkosFFT will not overwrite it.

yasahi-hpc:
I think it should work. Not sure about `hfft` and `ihfft`

PR #159, opened 2024-10-08
Rely on the user allocation in `ihfft`

tpadioleau:
Are we sure we want to discard the user allocation ?

### DIFF
--- a/fft/src/KokkosFFT_Transform.hpp
+++ b/fft/src/KokkosFFT_Transform.hpp
@@ -332,7 +332,7 @@ void ihfft(const ExecutionSpace& exec_space, const InViewType& in,
   OutViewType out_conj;
   rfft(exec_space, in, out, new_norm, axis, n);
   KokkosFFT::Impl::conjugate(exec_space, out, out_conj);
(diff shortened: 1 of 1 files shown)


tpadioleau:
linked to #158

yasahi-hpc (approved):
LGTM.
May be less performant than the original implementation?

tpadioleau:
> LGTM. May be less performant than the original implementation?

For now yes, I think the `conjugate` operation could be executed in-place ? We could get rid of the `out_conj` array.

yasahi-hpc:
> > LGTM. May be less performant than the original implementation?
> 
> For now yes, I think the `conjugate` operation could be executed in-place ? We could get rid of the `out_conj` array.

For this use case, we can make it in-place. For `hfft` with real input (not implemented yet), I need an out-place version of `conjugate` since it involves the data type conversion from `real` to `complex`. For this, I need to implement a helper to define a complex typed View from real typed View, like

```C++
View<double ***> -> View<Kokkos::complex<double> ***>
```

tpadioleau:
I see I agree

PR #161, opened 2024-10-09
Pass output views by const ref

### LINKED ISSUE #158: Consider passing output views by `const&`

tpadioleau:
closes #158

### DIFF
--- a/docs/api/hermitian/hfft.rst
+++ b/docs/api/hermitian/hfft.rst
@@ -5,4 +5,4 @@
 KokkosFFT::hfft
 ---------------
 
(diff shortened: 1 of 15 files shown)


yasahi-hpc (approved):
Codes are good. Could you please also update rst files under https://github.com/kokkos/kokkos-fft/tree/main/docs/api. The types for `doxygenfunction` are also need to be updated

yasahi-hpc (approved):

yasahi-hpc:
Thanks! I will merge this

PR #162, opened 2024-10-09
remove unnecessary barriers from examples

### LINKED ISSUE #157: Calling `exec.fence()` in the examples
How a reader should interpret the calls to `exec.fence()` in the examples ? As we pass `exec` to each function, I don't think they are required to have a valid execution ?

https://github.com/kokkos/kokkos-fft/blob/f64a013a8528d8a3bc552d4300105190e078d37a/examples/01_1DFFT/01_1DFFT.cpp#L25-L50

Should we get rid of them ?

yasahi-hpc:
Resolves #157


tpadioleau (approved):
Looks ok to me, I didn't see any host operation requiring a `fence()`.

PR #165, opened 2024-10-09
Make Plan class as a part of API

### LINKED ISSUE #156: Usage of the `KokkosFFT::Impl` namespace in example 06
In the example 06, we are currently using symbols from the `KokkosFFT::Impl` namespace, see
https://github.com/kokkos/kokkos-fft/blob/f64a013a8528d8a3bc552d4300105190e078d37a/examples/06_1DFFT_reuse_plans/06_1DFFT_reuse_plans.cpp#L31-L33

I think we should not show any usage of this implementation namespace in the examples.

I can think of different solutions:
- remove temporarily the example,
- move the symbols out of the `Impl` namespace if the feature is ready,
- move the symbols in an experimental namespace.

yasahi-hpc:
Resolves #156 

This PR aims at making Plan class as a part of API. 

- [x] `KokkosFFT::Impl::Plan` -> `KokkosFFT::Plan`
- [x] `KokkosFFT::Impl::fft_exec_impl` -> ~~`KokkosFFT::execute`~~ `plan.execute`
- [x] Remove unused variables from plan class `m_in_T` and `m_out_T`
- [x] Update implementation details and tests
- [x] Update [example 6](https://github.com/kokkos/kokkos-fft/tree/main/examples/06_1DFFT_reuse_plans)
- [x] Update docs 
- [x] Make internal types in Plan class private

### DIFF
--- a/docs/intro/using.rst
+++ b/docs/intro/using.rst
@@ -123,7 +123,7 @@ Reuse FFT plan
 --------------
 
 Apart from the basic APIs, Kokkos-fft offers the capability to create a FFT plan wrapping the FFT plans of backend libraries.
-We can reuse the FFT plan created once to perform FFTs multiple times on different data.
+We can reuse the FFT plan created once to perform FFTs multiple times on different data given that they have the same properties.
 In some backend, FFT plan creation leads to some overhead, wherein we need this functionality.
 (see :doc:`minimum working example<../samples/06_1DFFT_reuse_plans>`)
 
--- a/docs/intro/using.rst
+++ b/docs/intro/using.rst
@@ -122,8 +122,8 @@ The following listing shows examples with and without transpose operation.
 Reuse FFT plan
 --------------
 
-Apart from the basic APIs, Kokkos-fft APIs include overloaded APIs which can take a FFT plan as an argument.
-Using these overloaded APIs, we can reuse the FFT plan created before. 
+Apart from the basic APIs, Kokkos-fft offers the capability to create a FFT plan wrapping the FFT plans of backend libraries.
+We can reuse the FFT plan created once to perform FFTs multiple times on different data.
 In some backend, FFT plan creation leads to some overhead, wherein we need this functionality.
 (see :doc:`minimum working example<../samples/06_1DFFT_reuse_plans>`)
 
--- a/fft/src/KokkosFFT_Plans.hpp
+++ b/fft/src/KokkosFFT_Plans.hpp
@@ -15,34 +15,45 @@
 #include "KokkosFFT_default_types.hpp"
 #include "KokkosFFT_traits.hpp"
 #include "KokkosFFT_transpose.hpp"
+#include "KokkosFFT_normalization.hpp"
 #include "KokkosFFT_padding.hpp"
 #include "KokkosFFT_utils.hpp"
 
 #if defined(KOKKOS_ENABLE_CUDA)
 #include "KokkosFFT_Cuda_plans.hpp"
+#include "KokkosFFT_Cuda_transform.hpp"
 #ifdef ENABLE_HOST_AND_DEVICE
 #include "KokkosFFT_Host_plans.hpp"
+#include "KokkosFFT_Host_transform.hpp"
 #endif
 #elif defined(KOKKOS_ENABLE_HIP)
 #if defined(KOKKOSFFT_ENABLE_TPL_ROCFFT)
 #include "KokkosFFT_ROCM_plans.hpp"
+#include "KokkosFFT_ROCM_transform.hpp"
 #else
 #include "KokkosFFT_HIP_plans.hpp"
+#include "KokkosFFT_HIP_transform.hpp"
 #endif
 #ifdef ENABLE_HOST_AND_DEVICE
 #include "KokkosFFT_Host_plans.hpp"
+#include "KokkosFFT_Host_transform.hpp"
 #endif
 #elif defined(KOKKOS_ENABLE_SYCL)
 #include "KokkosFFT_SYCL_plans.hpp"
+#include "KokkosFFT_SYCL_transform.hpp"
 #ifdef ENABLE_HOST_AND_DEVICE
 #include "KokkosFFT_Host_plans.hpp"
+#include "KokkosFFT_Host_transform.hpp"
 #endif
 #elif defined(KOKKOS_ENABLE_OPENMP)
 #include "KokkosFFT_Host_plans.hpp"
+#include "KokkosFFT_Host_transform.hpp"
 #elif defined(KOKKOS_ENABLE_THREADS)
 #include "KokkosFFT_Host_plans.hpp"
+#include "KokkosFFT_Host_transform.hpp"
 #else
 #include "KokkosFFT_Host_plans.hpp"
+#include "KokkosFFT_Host_transform.hpp"
 #endif
 
 namespace KokkosFFT {
@@ -134,12 +145,6 @@ class Plan {
   //! extents of input/output views
   extents_type m_in_extents, m_out_extents;
   ///@}
-
-  //! @{
-  //! Internal buffers used for transpose
-  nonConstInViewType m_in_T;
-  nonConstOutViewType m_out_T;
-  //! @}
 
   //! Internal work buffer (for rocfft)
   BufferViewType m_buffer;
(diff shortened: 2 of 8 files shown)


tpadioleau:
I have questions about the API of `Plan`
- do you think all member functions of `KokkosFFT::Plan` need to be public ?
- are all member variables needed ? I don't think we use `m_out_T` ?
- will the member functions appear in the documentation ?

yasahi-hpc:
> I have questions about the API of `Plan`
> 
> * do you think all member functions of `KokkosFFT::Plan` need to be public ?

Actually, it may be better to make `KokkosFFT::execute` as a method of `KokkosFFT::Plan` class.
This way we can make most functions private. I do not remember why I did not do so.

> * are all member variables needed ? I don't think we use `m_out_T` ?

Sorry about that. These should be suppressed. We may add an option to use these variables to reuse the internal buffers in the future.

> * will the member functions appear in the documentation ?

I think we can make most of them private and add docstrings to public methods only.

yasahi-hpc:
@tpadioleau 
May I have your review please?
I think most of the issues are resolved

tpadioleau (changes requested):
I had a quick look. I think it is better now that we hide most of member functions of `KokkosFFT::Plan`.

My main concern on the current API is about `in` and `out` views. They are being passed both in the constructor and in `execute`. It looks overconstrained, it is not clear to me whether one can use different views in `execute` from views that were passed in the constructor. As far as I understand it is due to the FFTW API. Reading FFTW documentation, I understand we can use different arrays but with some constraints.

@cedricchevalier19 any opinion on that ?

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -271,6 +275,81 @@ class Plan {
...
   Plan(Plan&&)            = delete;
 
+  /// \brief Execute FFT on input and output Views with normalization
+  ///
+  /// \param in [in] Input data
+  /// \param out [out] Ouput data
+  template <typename InViewType2, typename OutViewType2>
+  void execute(
+      const InViewType2& in, const OutViewType2& out,
+      KokkosFFT::Normalization norm = KokkosFFT::Normalization::backward) {
```
tpadioleau, line 285:
```suggestion
  void execute(
      const InViewType2& in, const OutViewType2& out,
      KokkosFFT::Normalization norm = KokkosFFT::Normalization::backward) const {
```

Could it make sense to make the function `const` ?

yasahi-hpc:
> I had a quick look. I think it is better now that we hide most of member functions of `KokkosFFT::Plan`.
> 
> My main concern on the current API is about `in` and `out` views. They are being passed both in the constructor and in `execute`. It looks overconstrained, it is not clear to me whether one can use different views in `execute` from views that were passed in the constructor. As far as I understand it is due to the FFTW API. Reading FFTW documentation, I understand we can use different arrays but with some constraints.
> 
> @cedricchevalier19 any opinion on that ?

Thanks! 
`execute` interface is needed in order to reuse FFT plans.
Following FFTW, we allow use to change the internal data of views, but do not allow to change 
1. `Types` 
2. `extents`

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -271,6 +275,81 @@ class Plan {
   Plan& operator=(Plan&&) = delete;
   Plan(Plan&&)            = delete;
 
+  /// \brief Execute FFT on input and output Views with normalization
+  ///
+  /// \param in [in] Input data
+  /// \param out [out] Ouput data
+  template <typename InViewType2, typename OutViewType2>
```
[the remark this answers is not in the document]
yasahi-hpc: As you said, they are no longer needed. I have suppressed this.

fft/src/KokkosFFT_Plans.hpp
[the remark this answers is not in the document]
yasahi-hpc: Should have been public. 
Actually, some of the types are not necessary. I have suppressed them

tpadioleau, line 1: I would keep types private for now, let's introduce them when users ask for them ?

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -271,23 +269,86 @@ class Plan {
   Plan& operator=(Plan&&) = delete;
   Plan(Plan&&)            = delete;
 
+  /// \brief Execute FFT on input and output Views with normalization
+  ///
+  /// \param in [in] Input data
+  /// \param out [out] Ouput data
+  void execute(const InViewType& in, const OutViewType& out,
```
tpadioleau, line 276: documentation of `norm` seems to be missing.

fft/src/KokkosFFT_Plans.hpp
tpadioleau, line 1: I had a look at FFTW, I understood that the plan takes input and output arrays to find the best algorithms. If we use `FFTW_ESTIMATE`, then I think this API will only use the arrays to check if it will be an in-place algorithm ? Do you confirm ?

tpadioleau, line 1: I am not 100% convinced of templating the plan by Views. For example do we want to make a difference between plans using atomic views from non-atomic views ?

tpadioleau (commented):
Seems a good feature and well tested. I added general questions on our API

fft/src/KokkosFFT_Plans.hpp
yasahi-hpc: I agree. We no longer need to make them public. I made them private

fft/src/KokkosFFT_Plans.hpp
yasahi-hpc: Fixed accordingly.

fft/src/KokkosFFT_Plans.hpp
yasahi-hpc: > I had a look at FFTW, I understood that the plan takes input and output arrays to find the best algorithms. If we use FFTW_ESTIMATE, then I think this API will only use the arrays to check if it will be an in-place algorithm ? Do you confirm ?

I think the plan estimation is more than just checking in-place or not. It should at least factorize the size of FFTs (including pre-computation of twiddles) and analyze the dimension (how to apply FFTs over different dimensions). I expect FFTW is doing more than these

yasahi-hpc: > I am not 100% convinced of templating the plan by Views. For example do we want to make a difference between plans using atomic views from non-atomic views ?

As you pointed out, we cannot take into account memory traits since these traits cannot be transferred to FFT libraries. Still, I think using View meta data is the most practical way to construct a plan.

tpadioleau: I totally agree, I meant regarding the pointers `in` and `out`.

docs/intro/using.rst
```diff
@@ -122,8 +122,8 @@ The following listing shows examples with and without transpose operation.
 Reuse FFT plan
 --------------
 
-Apart from the basic APIs, Kokkos-fft APIs include overloaded APIs which can take a FFT plan as an argument.
-Using these overloaded APIs, we can reuse the FFT plan created before. 
+Apart from the basic APIs, Kokkos-fft offers the capability to create a FFT plan wrapping the FFT plans of backend libraries.
+We can reuse the FFT plan created once to perform FFTs multiple times on different data.
```
tpadioleau, line 126:
```suggestion
One can reuse the FFT plan created once to perform FFTs multiple times on different data given that they have the same properties.
```
Just a word to say we cannot pass anything ?

fft/src/KokkosFFT_Plans.hpp
tpadioleau: Ok let's start with this API. I think we can discuss that with Damien on a monday's meeting.

tpadioleau (approved):
Ok for me! let's try that

fft/src/KokkosFFT_Plans.hpp
yasahi-hpc: Sure

yasahi-hpc: We will start analyze the pointers with in-place capability. But we do not modify them in plan creation

docs/intro/using.rst
yasahi-hpc: Fixed accordingly

yasahi-hpc:
@tpadioleau 
Thank you for reviews.
If you feel good, I will merge this

tpadioleau (approved):

Issue #170, opened 2024-10-11
Implement conversion helpers between `Kokkos::Array` and `std::array`


tpadioleau:
We have many of these conversions:
https://github.com/kokkos/kokkos-fft/blob/4fe0f2e5e2247904bcf35fcd2e5edbc2a081bd4a/common/src/KokkosFFT_transpose.hpp#L156
https://github.com/kokkos/kokkos-fft/blob/4fe0f2e5e2247904bcf35fcd2e5edbc2a081bd4a/common/src/KokkosFFT_transpose.hpp#L190
https://github.com/kokkos/kokkos-fft/blob/4fe0f2e5e2247904bcf35fcd2e5edbc2a081bd4a/common/src/KokkosFFT_transpose.hpp#L227
...

It would be helpful to implement once and for all a helper to go back and forth between `Kokkos::Array` and `std::array`. It could even be an upstream proposal for Kokkos core.

dalg24:
https://godbolt.org/z/3vfv34MK3

tpadioleau:
@dalg24 do you think this can go to Kokkos 4.5 ?

dalg24:
I am torn on this.  I don't mind too much the `to_array` free functions but I am afraid others will soon come and ask for regular conversion operations and I am strongly against these.
We already have some of that mess in Kokkos::complex in the name of "compatibility".

tpadioleau:
I would be happy to have these free functions, even for other projects. That could also be seen as helper tools in porting existing C++ codes to GPU.

About conversion operations we could argue that we only want to support at most the API of `std::array`.

dalg24:
`Kokkos::to_array(std::array) -> Kokkos::Array` looks like the obvious choice but what did you have in mind for the conversion Kokkos:: to std::array?
`Kokkos::to_std_array(Kokkos::Array) -> std::array`?
Apparently Kokkos::pair has a to_std_pair() member function
https://github.com/kokkos/kokkos/blob/abcb3dce75effd5cd7a2d73e9e51a9202573acab/core/src/Kokkos_Pair.hpp#L144-L146

tpadioleau:
Yes seems fine
```cpp
namespace Kokkos
{
    template <class T, std::size_t N>
    using array = Kokkos::Array<T, N>;

    template <class T, std::size_t N>
    array<T, N> to_array(std::array<T, N> const&/&& x)

    template <class T, std::size_t N>
[...]

Issue #85, opened 2024-03-11
support plan: latest two releases


yasahi-hpc:
- [x] Update a unit-test with new `Kokkos::Array` once we update minimum Kokkos version to 4.4 or later (see #170)

yasahi-hpc:
Kokkos now released [v.4.3.0](https://github.com/kokkos/kokkos/releases/tag/4.3.00). 
To support latest two leases, should we add CI for v.4.2 and v.4.3 in matrix?

yasahi-hpc:
I think this has been resolved

PR #171, opened 2024-10-12
Apply to_array helper

yasahi-hpc:
Improves #170 

This PR aims at introducing [`to_array`](https://godbolt.org/z/3vfv34MK3) helper which is used in `Transpose` functions. 
For the compile time testing, we need a newer version of Kokkos.

### DIFF
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -226,11 +226,11 @@ constexpr Kokkos::Array<T, N> to_array_rvalue_helper(
   return {{std::move(a[Is])...}};
 }
 
-template <class T, size_t N, size_t... Is>
+template <class T, size_t N>
 constexpr Kokkos::Array<T, N> to_array(const std::array<T, N>& a) {
   return to_array_lvalue_helper(a, std::make_index_sequence<N>());
 }
-template <class T, size_t N, size_t... Is>
+template <class T, size_t N>
 constexpr Kokkos::Array<T, N> to_array(std::array<T, N>&& a) {
   return to_array_rvalue_helper(std::move(a), std::make_index_sequence<N>());
 }
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -217,7 +217,7 @@ Layout create_layout(const std::array<int, N>& extents) {
 
 template <class T, size_t N, size_t... Is>
 constexpr Kokkos::Array<T, N> to_array_lvalue_helper(
-    std::array<T, N>& a, std::index_sequence<Is...>) {
+    const std::array<T, N>& a, std::index_sequence<Is...>) {
   return {{a[Is]...}};
 }
 template <class T, size_t N, size_t... Is>
@@ -227,7 +227,7 @@ constexpr Kokkos::Array<T, N> to_array_rvalue_helper(
 }
 
 template <class T, size_t N, size_t... Is>
-constexpr Kokkos::Array<T, N> to_array(std::array<T, N>& a) {
+constexpr Kokkos::Array<T, N> to_array(const std::array<T, N>& a) {
   return to_array_lvalue_helper(a, std::make_index_sequence<N>());
 }
 template <class T, size_t N, size_t... Is>
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -215,6 +215,26 @@ Layout create_layout(const std::array<int, N>& extents) {
   return layout;
 }
 
+template <class T, size_t N, size_t... Is>
+constexpr Kokkos::Array<T, N> to_array_lvalue_helper(
+    std::array<T, N>& a, std::index_sequence<Is...>) {
+  return {{a[Is]...}};
+}
+template <class T, size_t N, size_t... Is>
+constexpr Kokkos::Array<T, N> to_array_rvalue_helper(
+    std::array<T, N>&& a, std::index_sequence<Is...>) {
+  return {{std::move(a[Is])...}};
+}
+
+template <class T, size_t N, size_t... Is>
+constexpr Kokkos::Array<T, N> to_array(std::array<T, N>& a) {
+  return to_array_lvalue_helper(a, std::make_index_sequence<N>());
+}
+template <class T, size_t N, size_t... Is>
+constexpr Kokkos::Array<T, N> to_array(std::array<T, N>&& a) {
+  return to_array_rvalue_helper(std::move(a), std::make_index_sequence<N>());
+}
+
 }  // namespace Impl
 }  // namespace KokkosFFT
 
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -215,22 +215,22 @@ Layout create_layout(const std::array<int, N>& extents) {
   return layout;
 }
 
-template <class T, size_t N, size_t... Is>
+template <class T, std::size_t N, std::size_t... Is>
 constexpr Kokkos::Array<T, N> to_array_lvalue_helper(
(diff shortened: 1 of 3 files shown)


common/src/KokkosFFT_utils.hpp
```diff
@@ -215,6 +215,26 @@ Layout create_layout(const std::array<int, N>& extents) {
...
+}
+
+template <class T, size_t N, size_t... Is>
+constexpr Kokkos::Array<T, N> to_array(std::array<T, N>& a) {
+  return to_array_lvalue_helper(a, std::make_index_sequence<N>());
+}
+template <class T, size_t N, size_t... Is>
+constexpr Kokkos::Array<T, N> to_array(std::array<T, N>&& a) {
+  return to_array_rvalue_helper(std::move(a), std::make_index_sequence<N>());
+}
```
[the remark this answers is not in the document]
yasahi-hpc: Makes sense. I have fixed accordingly

common/src/KokkosFFT_utils.hpp
```diff
@@ -215,6 +215,26 @@ Layout create_layout(const std::array<int, N>& extents) {
...
+constexpr Kokkos::Array<T, N> to_array_rvalue_helper(
+    std::array<T, N>&& a, std::index_sequence<Is...>) {
+  return {{std::move(a[Is])...}};
+}
+
+template <class T, size_t N, size_t... Is>
+constexpr Kokkos::Array<T, N> to_array(const std::array<T, N>& a) {
+  return to_array_lvalue_helper(a, std::make_index_sequence<N>());
+}
+template <class T, size_t N, size_t... Is>
```
dalg24, line 233:
```suggestion
template <class T, size_t N>
constexpr Kokkos::Array<T, N> to_array(const std::array<T, N>& a) {
  return to_array_lvalue_helper(a, std::make_index_sequence<N>());
}
template <class T, size_t N>
```

yasahi-hpc:
We may need a newer version of Kokkos for nicer tests. Should we update Kokkos to 4.4 and update docs accordingly?

tpadioleau (commented):
We can try

common/unit_test/Test_Utils.cpp
```diff
@@ -541,3 +541,17 @@ TEST(IndexSequence, 3Dto5D) {
...
+}
+
+TEST(ToArray, rvalue) {
+  ASSERT_EQ(KokkosFFT::Impl::to_array(std::array{1, 2}),
+            (Kokkos::Array<int, 2>{1, 2}));
+  // [TO DO] Need to update Kokkos version for compile time test
+  // GTEST_SKIP() << "Skipping ToArray::rvalue compile time test";
+  // static_assert(KokkosFFT::Impl::to_array(std::array{1, 2}) ==
+  // Kokkos::Array{1, 2});
+}
```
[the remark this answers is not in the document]
yasahi-hpc: I fixed the tests to be compatible with Kokkos 4.2

common/src/KokkosFFT_utils.hpp
```diff
@@ -215,6 +215,26 @@ Layout create_layout(const std::array<int, N>& extents) {
...
+}
+
+template <class T, size_t N>
+constexpr Kokkos::Array<T, N> to_array(const std::array<T, N>& a) {
+  return to_array_lvalue_helper(a, std::make_index_sequence<N>());
+}
+template <class T, size_t N>
+constexpr Kokkos::Array<T, N> to_array(std::array<T, N>&& a) {
+  return to_array_rvalue_helper(std::move(a), std::make_index_sequence<N>());
+}
```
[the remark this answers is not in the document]
yasahi-hpc: I have fixed accordingly

tpadioleau (approved):
LGTM

common/unit_test/Test_Utils.cpp
```diff
@@ -541,3 +541,24 @@ TEST(IndexSequence, 3Dto5D) {
...
+TEST(ToArray, rvalue) {
+  Kokkos::Array<int, 3> ref_arr{1, 2, 3};
+  auto kokkos_arr = KokkosFFT::Impl::to_array(std::array{1, 2, 3});
+  for (std::size_t i = 0; i < ref_arr.size(); ++i) {
+    ASSERT_EQ(kokkos_arr[i], ref_arr[i]);
+  }
+  // [TO DO] Need to update Kokkos version for compile time test
+  // GTEST_SKIP() << "Skipping ToArray::rvalue compile time test";
+  // static_assert(KokkosFFT::Impl::to_array(std::array{1, 2}) ==
+  // Kokkos::Array{1, 2});
```
tpadioleau, line 563: If you want,  you can already implement the test using the comparison operator inside `ifdef` based on the version of Kokkos ?

yasahi-hpc: If you are fine, I would like to keep it in comments since we do not test that in CI for the moment.

tpadioleau: Yes I am ok with that. One question though, how do we keep track that we will need to update this snippet when we increase the minimum version of Kokkos ?

yasahi-hpc: What about tracking in an [issue](#85)?

tpadioleau: I am not really confident on remembering that there is actually an issue about this when we bump the minimum version. We can try.

yasahi-hpc: Let me update the internal Kokkos version. It would be the simplest way

yasahi-hpc: I have changed my mind to follow the Kokkos convention. https://github.com/kokkos/kokkos/blob/abcb3dce75effd5cd7a2d73e9e51a9202573acab/core/src/Kokkos_Array.hpp#L388-L396

What do you think?

common/unit_test/Test_Utils.cpp
```diff
@@ -541,3 +541,14 @@ TEST(IndexSequence, 3Dto5D) {
...
 }
+
+TEST(ToArray, lvalue) {
+  std::array<int, 3> arr{1, 2, 3};
+  ASSERT_EQ(KokkosFFT::Impl::to_array(arr), (Kokkos::Array{1, 2, 3}));
+}
+
+TEST(ToArray, rvalue) {
+  GTEST_SKIP() << "Skipping ToArray::rvalue compile time test";
+  static_assert(KokkosFFT::Impl::to_array(std::array{1, 2}) ==
```
tpadioleau, line 552: I am not sure we need to make it a compile time test ? We only do that for rvalue and not lvalue ?

I don't think we need to be as general as we would be inside Kokkos core, runtime tests should be enough to cover our needs. I would even consider removing the `constexpr` annotation on `to_array` ?

yasahi-hpc: I have modified the test accordingly.
As for the `constexpr` annotation, I would like to follow the `Kokkos` convention since it is likely to be an official implementation

common/unit_test/Test_Utils.cpp
```diff
@@ -541,3 +541,12 @@ TEST(IndexSequence, 3Dto5D) {
   EXPECT_EQ(default_axes1, ref_axes1);
   EXPECT_EQ(default_axes2, ref_axes2);
 }
+
+TEST(ToArray, lvalue) {
+  std::array<int, 3> arr{1, 2, 3};
```
tpadioleau, line 546:
```suggestion
  std::array arr{1, 2, 3};
```
Just to be uniform

tpadioleau (approved):

common/unit_test/Test_Utils.cpp
tpadioleau: Ok for me

common/unit_test/Test_Utils.cpp
yasahi-hpc: Fixed accordingly

tpadioleau (approved):

PR #172, opened 2024-10-13
Remove maybe_unused to actually used variables

tpadioleau:


yasahi-hpc:
Actually, we need them for [nvcc](https://github.com/kokkos/kokkos-fft/actions/runs/11312483924/job/31459949881?pr=172) to avoid warnings. I thought these warnings lead to the compiler errors in CI. I will fix the CI.

tpadioleau:
> Actually, we need them for [nvcc](https://github.com/kokkos/kokkos-fft/actions/runs/11312483924/job/31459949881?pr=172) to avoid warnings. I thought these warnings lead to the compiler errors in CI. I will fix the CI.

Oh I see, let me try removing a pair of brackets

yasahi-hpc:
> > Actually, we need them for [nvcc](https://github.com/kokkos/kokkos-fft/actions/runs/11312483924/job/31459949881?pr=172) to avoid warnings. I thought these warnings lead to the compiler errors in CI. I will fix the CI.
> 
> Oh I see, let me try removing a pair of brackets

Thanks!
For some reason, these warnings do not cause compile errors even with `-Werror` and `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON`.

common/src/KokkosFFT_layouts.hpp
```diff
@@ -30,10 +30,8 @@ auto get_extents(const InViewType& in, const OutViewType& out,
                      "input axes are not valid for the view");
 
   constexpr std::size_t rank = InViewType::rank;
-  [[maybe_unused]] int inner_most_axis =
-      std::is_same_v<array_layout_type, typename Kokkos::LayoutLeft>
-          ? 0
-          : (rank - 1);
+  int inner_most_axis =
+      std::is_same_v<array_layout_type, Kokkos::LayoutLeft> ? 0 : (rank - 1);
```
yasahi-hpc, line 34: This variable `inner_most_axis` is actually unused for `C2C` transform which is identified at compile time. 
I thought in this case compilers regard this variable unused. 
However, I found no compiler errors

yasahi-hpc (approved):

common/src/KokkosFFT_layouts.hpp
tpadioleau: Could be because there is no `if constexpr` ?

yasahi-hpc: You are right. lines 64 and 73 must be `if constexpr`.

yasahi-hpc: If you are fine, I will merge this and fix `if constexpr` in another PR.

tpadioleau: Yes, you might need `[[maybe_unused]]` then.

yasahi-hpc: That is exactly what I had in mind

PR #173, opened 2024-10-13
Fix inconsistent IndexType

tpadioleau:
An alternative fix is to use `IndexType<int>`


yasahi-hpc (approved):
LGTM. I will merge this

PR #174, opened 2024-10-14
Use if constexpr for compile time if conditional

yasahi-hpc:
In this PR, we use `if constexpr` for compile time if conditional.


tpadioleau (approved):
Ok for me, I compiled withtout `[[maybe_unused]]` and did not get any warning :D compilers...

yasahi-hpc:
> Ok for me, I compiled withtout `[[maybe_unused]]` and did not get any warning :D compilers...

It is a bit surprising. Should we remove that then?

tpadioleau:
> Should we remove that then?

No I think it is legitimate to keep it, there is a compile-time path where the variable is not used.

yasahi-hpc:
Thanks for comments. I will merge this.

PR #179, opened 2024-10-24
Avoid variables starting from underscore under common

yasahi-hpc:
Improves #81

- [x] Avoid variables starting from under score under `common/src`
- [x] Avoid variables starting from under score under `common/unit_test`


tpadioleau:
It is not easy to review, seems ok to me

tpadioleau (approved):

yasahi-hpc:
> It is not easy to review, seems ok to me

Fully agree.
Thank you for your review.

After finishing I will add this to clangitdy 
https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html

PR #181, opened 2024-10-28
Harmonize the library names as kokkos-fft

yasahi-hpc:
This PR aims at harmonizing the library name as `kokkos-fft`. 
Because of the Copyright on the top of each file, this PR seems extremely big. 
So, I would like to finish this first and update the docs in the other PR.


yasahi-hpc:
Close this. I will redo this from original repo

PR #188, opened 2024-11-01
Update workflow dependencies

pzehner:
Update actions to their latest version.

### DIFF
--- a/.github/workflows/cleanup_base.yaml
+++ b/.github/workflows/cleanup_base.yaml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: Cleanup old images
-        uses: SmartsquareGmbH/delete-old-packages@v0.7.0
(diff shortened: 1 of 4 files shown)


.github/workflows/cleanup_base.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: Cleanup old images
-        uses: SmartsquareGmbH/delete-old-packages@v0.7.0
+        uses: SmartsquareGmbH/delete-old-packages@v0.8
```
tpadioleau, line 21:
```suggestion
        uses: SmartsquareGmbH/delete-old-packages@v0.8.1
```

It seems that the current policy is to fix also the minor version

tpadioleau (approved):
Ok for me

tpadioleau (approved):
lgtm

PR #189, opened 2024-11-01
Improve tests using Singularity

pzehner:
It seems now that GHCR requires users to be logged to pull images. This PR fixes it for the CI part on Ruche.


pzehner:
In case, I login Docker before pulling images, even if it was already working (maybe the environment is prepared?).

PR #190, opened 2024-11-04
Add inplace transform example

yasahi-hpc:
This PR aims to adding an example for inplace 2D transforms. The documentation should be updated in another PR.


examples/08_inplace_FFT/08_inplace_FFT.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    shape_type<2> shape;
+    shape[0] = n0;
+    shape[1] = n1;
+
+    // Forward and backward complex to complex transform
+    // Define a 2D complex view to handle data
+    RightView2D<Kokkos::complex<double>> xc2c("xc2c", n0, n1);
+
+    // Create an unmanaged view on the same data
```
tpadioleau, line 34:
```suggestion
    // Create unmanaged views on the same data
```

examples/08_inplace_FFT/08_inplace_FFT.cpp
```diff
@@ -0,0 +1,88 @@
...
+    RightView2D<Kokkos::complex<double>> xc2c_hat(xc2c.data(), n0, n1),
```
tpadioleau, line 35: Actually for the complex to complex transform, we don't require unmanaged views:
- should we make them managed views ?
- be consistent with the c2r/r2c case ?

examples/08_inplace_FFT/08_inplace_FFT.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+int main(int argc, char *argv[]) {
+  Kokkos::initialize(argc, argv);
+  {
+    constexpr int n0 = 128, n1 = 128;
+    const Kokkos::complex<double> z(1.0, 1.0);
+
+    shape_type<2> shape;
+    shape[0] = n0;
+    shape[1] = n1;
```
tpadioleau, line 28: Can we initialize `shape{n0, n1}` ?

examples/08_inplace_FFT/08_inplace_FFT.cpp
```diff
@@ -0,0 +1,88 @@
...
+                                 n1);
+
+    // Fill the input view with random data in complex space
+    Kokkos::fill_random(exec, xc2r, random_pool, z);
+
+    // Perform the complex to real transform
+    // [Important] You must use xc2r_hat to define the FFT shape correctly
+    KokkosFFT::irfft2(exec, xc2r, xc2r_hat);
+
+    Kokkos::fence();
```
tpadioleau, line 83:
```suggestion
    exec.fence();
```

examples/08_inplace_FFT/08_inplace_FFT.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+template <std::size_t DIM>
+using axis_type = KokkosFFT::axis_type<DIM>;
+template <std::size_t DIM>
+using shape_type = KokkosFFT::shape_type<DIM>;
+
+int main(int argc, char *argv[]) {
+  Kokkos::initialize(argc, argv);
+  {
+    constexpr int n0 = 128, n1 = 128;
```
tpadioleau, line 23: Not only this example but I am not convinced of making `n0` and `n1` `constexpr`. It makes me wonder if kokkos-fft can take advantage of it or not ?

examples/08_inplace_FFT/08_inplace_FFT.cpp
```diff
@@ -0,0 +1,88 @@
...
+    // Real to complex transform
+    // Define a 2D complex view to handle data
+    RightView2D<Kokkos::complex<double>> xr2c_hat("xr2c", n0, n1 / 2 + 1);
+
+    // Create an unmanaged view on the same data with the FFT shape,
+    // that is (n0, n1) -> (n0, n1/2+1) R2C transform
+    // The shape is incorrect from the view point of casting to real
+    // For casting, the shape should be (n0, (n0/2+1) * 2)
+    RightView2D<double> xr2c(reinterpret_cast<double *>(xr2c_hat.data()), n0,
+                             n1),
```
tpadioleau, line 55: Should we provide helper functions to avoid users using `reinterpret_cast` ?

tpadioleau (dismissed):
Looks almost good to me, just minor comments

examples/08_inplace_FFT/08_inplace_FFT.cpp
yasahi-hpc: That is a good point. 
I was just thinking of consistency with c2r and r2c.
Probably better to reuse views for simplicity.

examples/08_inplace_FFT/08_inplace_FFT.cpp
yasahi-hpc: True. I will fix that

examples/08_inplace_FFT/08_inplace_FFT.cpp
yasahi-hpc: Sure

examples/08_inplace_FFT/08_inplace_FFT.cpp
yasahi-hpc: No. Just `const` is good enough

examples/08_inplace_FFT/08_inplace_FFT.cpp
yasahi-hpc: That may be ideal. 
Can I make it another PR?

examples/08_inplace_FFT/08_inplace_FFT.cpp
yasahi-hpc: Actually, this variable is unnecessary. I have removed this

common/src/KokkosFFT_Extents.hpp
```diff
@@ -33,7 +33,7 @@ namespace Impl {
 template <typename InViewType, typename OutViewType, std::size_t DIM = 1>
 auto get_extents(const InViewType& in, const OutViewType& out,
                  axis_type<DIM> axes, shape_type<DIM> shape = {},
-                 bool is_inplace = false) {
+                 [[maybe_unused]] bool is_inplace = false) {
```
tpadioleau, line 36: Why does it become needed only in this PR ?

examples/08_inplace_FFT/08_inplace_FFT.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+using execution_space = Kokkos::DefaultExecutionSpace;
+
+template <typename T>
+using RightView2D = Kokkos::View<T **, Kokkos::LayoutRight, execution_space>;
+
+template <std::size_t DIM>
+using axis_type = KokkosFFT::axis_type<DIM>;
+template <std::size_t DIM>
+using shape_type = KokkosFFT::shape_type<DIM>;
```
tpadioleau, line 18:
```suggestion
```

tpadioleau (changes requested):

examples/08_inplace_FFT/08_inplace_FFT.cpp
tpadioleau: This was not detected by the warnings as errors ?

tpadioleau:
Out of curiosity, I guess there is no equivalent in scipy ? It does not support inplace transformation ?

common/src/KokkosFFT_Extents.hpp
yasahi-hpc: Actually, we should have encountered a compiler error due to the unused variable for C2C transform. 
For some reason, we did not get compiler errors before introducing the inplace C2C transform with the same `View`.

examples/08_inplace_FFT/08_inplace_FFT.cpp
yasahi-hpc: Sorry, I forgot to remove this as well

common/src/KokkosFFT_Extents.hpp
tpadioleau: Unexpected indeed, but I agree that it should be added anyway.

yasahi-hpc:
@tpadioleau Thank you for your approval. I will merge this

PR #193, opened 2024-11-04
allow to continue run tests even if ruche CI fails

yasahi-hpc:
In this RP, we allow to continue run tests even if ruche CI fails.

### DIFF
--- a/.github/workflows/build_test.yaml
+++ b/.github/workflows/build_test.yaml
@@ -313,7 +313,6 @@ jobs:
     strategy:
       matrix:
         backend:
-          unstable: [false]
           # run CUDA tests on Ruche supercomputer with Singularity
           - name: cuda
             image: nvcc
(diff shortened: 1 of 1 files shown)


tpadioleau:
I still think it is worth it, we should try to make it work at some point

.github/workflows/build_test.yaml
```diff
@@ -310,13 +311,16 @@ jobs:
...
+      fail-fast: false
       matrix:
+        unstable: [false]
         backend:
           # run CUDA tests on Ruche supercomputer with Singularity
           - name: cuda
             image: nvcc
             runner: [self-hosted, cuda]
             use_singularity: true
+            unstable: true
```
tpadioleau, line 323: Was it meant to override `matrix.unstable` ? If I understand correctly this creates a new variable `matrix.backend.unstable` ?

I think we need an entry for each backend ?

tpadioleau (changes requested):

.github/workflows/build_test.yaml
yasahi-hpc: Thank you for the comments.

Ah, that is true. We need to introduce `matrix.backend.unstable` rather than `matrix.unstable`

tpadioleau (approved):
Ok let's try!

yasahi-hpc:
@tpadioleau Thanks again for the review.
Fortunately or unfortunately, ruche CI worked this time. 
Hopefully after merging this, CI continues even if ruche CI fails

PR #196, opened 2024-11-05
Add missing checks

yasahi-hpc:
This PR aims at adding missing checks.

- [x] Add extents consistency check for all-extents. Previously, we checked only for the transform axis of `R2C` and `C2R` transforms.
- [x] Disallow to use inplace Plan on out-of-place Views (plan is not identical)
- [X] Add throw tests for these cases

### DIFF
--- a/common/src/KokkosFFT_Extents.hpp
+++ b/common/src/KokkosFFT_Extents.hpp
@@ -75,6 +75,22 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   static_assert(!(is_real_v<in_value_type> && is_real_v<out_value_type>),
                 "get_extents: real to real transform is not supported");
 
+  for (std::size_t i = 0; i < rank; i++) {
+    // The requirement for inner_most_axis is different for transform type
+    if (static_cast<int>(i) == inner_most_axis) continue;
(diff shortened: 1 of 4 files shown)


tpadioleau (dismissed):
ok for me, still working fine on GPU ?

yasahi-hpc:
> ok for me, still working fine on GPU ?

Thanks! 
I will fix accordingly. 
It runs correctly on GPUs. The current failure stems from the time-out issue

tpadioleau:
> It runs correctly on GPUs. The current failure stems from the time-out issue

Strange, are you sure something is not broken ? I would expect that within a day a single GPU would be available.

tpadioleau:
I ran `squeue` on the supercomputer and it returns an empty list of jobs.

yasahi-hpc:
> I ran `squeue` on the supercomputer and it returns an empty list of jobs.

That is true. Let me check

yasahi-hpc:
> > I ran `squeue` on the supercomputer and it returns an empty list of jobs.
> 
> That is true. Let me check

I found that the runner is offline

yasahi-hpc:
Seems OK now. I will merge this

PR #205, opened 2024-12-02
Add derivative example

### LINKED ISSUE #49: Write a small example of numpy fft derivative

yasahi-hpc:
Resolves #49


examples/09_derivative/09_derivative.cpp
```diff
@@ -0,0 +1,222 @@
...
+  Kokkos::parallel_for(
+      "ComputeDerivative", range2d, KOKKOS_LAMBDA(const int iy, const int ix) {
+        auto ikx_tmp = ikx(iy, ix), iky_tmp = iky(iy, ix);
+        for (int iz = 0; iz < nz; ++iz) {
+          u_hat(iz, iy, ix) =
+              (ikx_tmp * u_hat(iz, iy, ix) + iky_tmp * u_hat(iz, iy, ix));
+        }
+      });
+
+  // Baclward transform u_hat -> u (=IFFT (u_hat))
```
rb214678, line 183: There is a small typo on "Baclward".

yasahi-hpc:
Thank you for the review.

> When comparing the derivative of a function using FFT-based methods and the one computed with the analytical solution, in the Python script you use `np.testing.assert_allclose` with a tolerance. In the C++ code you do it "by hand" with some loop. Do you need these checks to be identical ? If yes it is the case ?

They are not exactly the same. 
I am currently considering to introduce an additional package for testing like numpy. 
Once I introduce it, I can use that for the sanity check in this example.
However, it is out of the scope of this PR, so I keep it as is.

rb214678 (approved):

PR #208, opened 2024-12-03
Fix RAII issue by introducing wrapper classes for backend plans

### LINKED ISSUE #197: RAII issue
I suspect potential memory leaks.

Consider the snippet https://github.com/kokkos/kokkos-fft/blob/4582878070fbdd93d13e1235d4aaa37ea5d4d794/fft/src/KokkosFFT_Cuda_plans.hpp#L39-L54

It is called inside the `Plan` constructor. Let's assume the second `KOKKOSFFT_THROW_IF` actually throws, the backend `destroy_plan_and_info` function will never be called because the `Plan` object will never finish construction.

I think we need to adhere to the RAII so that if any exception is thrown the backend plan object gets automatically freed.

yasahi-hpc:
Resolves #197 

In this PR, we try to resolve RAII issues by introducing a thin wrapper for backend fft plans.
Following modifications are made,

- [x] Introducing a `KokkosFFT_FFTW_Types.hpp` file including the `FFTW` plan wrapper class and common code bases related fftw
- [x] Introducing a wrapper class for automatic deallocation. Wrappers are added for `FFTW`, `cufft`, `hipfft`, and `rocfft` (`Scoped***Plan`) 
For CUDA, we have

```C++
struct ScopedCufftPlanType {
   cufftHandle m_plan;
   // Should we raise an error here, if it fails?
   ScopedCufftPlanType() { cufftCreate(&m_plan); }
   
   // Should we abort here, if it fails?
   ~ScopedCufftPlanType() { cufftDestroy(m_plan); }

   cufftHandle &plan() { return m_plan; }
};
```

- [x] `Info` and `Buffer` that are required only for rocfft are moved inside `rocfft` helper class
- [x] Add missing `cleanup_threads()` in the destructor of FFTW helper class
- [x] Remove `destroy_plan_and_info` since it is automatically done in destructors of helper classes

Questions
1. We are creating plans inside a constructor for `cufft` and `hipfft`. Should we raise assertions in constructor if it fails?
2. What should we do if destroy functions in destructor fails. Should we abort?

### DIFF
--- a/fft/src/KokkosFFT_Cuda_transform.hpp
+++ b/fft/src/KokkosFFT_Cuda_transform.hpp
@@ -10,45 +10,47 @@
 
 namespace KokkosFFT {
 namespace Impl {
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftReal* idata, cufftComplex* odata,
-                      int /*direction*/, Args...) {
-  cufftResult cufft_rt = cufftExecR2C(plan, idata, odata);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftReal* idata,
+                      cufftComplex* odata, int /*direction*/) {
+  cufftResult cufft_rt = cufftExecR2C(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecR2C failed");
 }
 
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftDoubleReal* idata,
-                      cufftDoubleComplex* odata, int /*direction*/, Args...) {
-  cufftResult cufft_rt = cufftExecD2Z(plan, idata, odata);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftDoubleReal* idata,
+                      cufftDoubleComplex* odata, int /*direction*/) {
+  cufftResult cufft_rt = cufftExecD2Z(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecD2Z failed");
 }
 
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftComplex* idata, cufftReal* odata,
-                      int /*direction*/, Args...) {
-  cufftResult cufft_rt = cufftExecC2R(plan, idata, odata);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftComplex* idata,
+                      cufftReal* odata, int /*direction*/) {
+  cufftResult cufft_rt = cufftExecC2R(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecC2R failed");
 }
 
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftDoubleComplex* idata,
-                      cufftDoubleReal* odata, int /*direction*/, Args...) {
-  cufftResult cufft_rt = cufftExecZ2D(plan, idata, odata);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftDoubleComplex* idata,
+                      cufftDoubleReal* odata, int /*direction*/) {
+  cufftResult cufft_rt = cufftExecZ2D(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecZ2D failed");
 }
 
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftComplex* idata,
-                      cufftComplex* odata, int direction, Args...) {
-  cufftResult cufft_rt = cufftExecC2C(plan, idata, odata, direction);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftComplex* idata,
+                      cufftComplex* odata, int direction) {
+  cufftResult cufft_rt =
+      cufftExecC2C(scoped_plan.plan(), idata, odata, direction);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecC2C failed");
 }
 
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftDoubleComplex* idata,
-                      cufftDoubleComplex* odata, int direction, Args...) {
-  cufftResult cufft_rt = cufftExecZ2Z(plan, idata, odata, direction);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftDoubleComplex* idata,
+                      cufftDoubleComplex* odata, int direction) {
+  cufftResult cufft_rt =
+      cufftExecZ2Z(scoped_plan.plan(), idata, odata, direction);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecZ2Z failed");
 }
 }  // namespace Impl
--- a/fft/src/KokkosFFT_Cuda_transform.hpp
+++ b/fft/src/KokkosFFT_Cuda_transform.hpp
@@ -10,45 +10,45 @@
 
 namespace KokkosFFT {
 namespace Impl {
-template <typename ScopedPlanType>
-inline void exec_plan(ScopedPlanType& scoped_plan, cufftReal* idata,
+
+struct ScopedCufftPlan;
+
+inline void exec_plan(const ScopedCufftPlan& scoped_plan, cufftReal* idata,
                       cufftComplex* odata, int /*direction*/) {
   cufftResult cufft_rt = cufftExecR2C(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecR2C failed");
 }
 
-template <typename ScopedPlanType>
-inline void exec_plan(ScopedPlanType& scoped_plan, cufftDoubleReal* idata,
-                      cufftDoubleComplex* odata, int /*direction*/) {
+inline void exec_plan(const ScopedCufftPlan& scoped_plan,
+                      cufftDoubleReal* idata, cufftDoubleComplex* odata,
+                      int /*direction*/) {
   cufftResult cufft_rt = cufftExecD2Z(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecD2Z failed");
 }
 
-template <typename ScopedPlanType>
-inline void exec_plan(ScopedPlanType& scoped_plan, cufftComplex* idata,
+inline void exec_plan(const ScopedCufftPlan& scoped_plan, cufftComplex* idata,
                       cufftReal* odata, int /*direction*/) {
   cufftResult cufft_rt = cufftExecC2R(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecC2R failed");
 }
 
-template <typename ScopedPlanType>
-inline void exec_plan(ScopedPlanType& scoped_plan, cufftDoubleComplex* idata,
-                      cufftDoubleReal* odata, int /*direction*/) {
+inline void exec_plan(const ScopedCufftPlan& scoped_plan,
+                      cufftDoubleComplex* idata, cufftDoubleReal* odata,
+                      int /*direction*/) {
   cufftResult cufft_rt = cufftExecZ2D(scoped_plan.plan(), idata, odata);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecZ2D failed");
 }
 
-template <typename ScopedPlanType>
-inline void exec_plan(ScopedPlanType& scoped_plan, cufftComplex* idata,
+inline void exec_plan(const ScopedCufftPlan& scoped_plan, cufftComplex* idata,
                       cufftComplex* odata, int direction) {
   cufftResult cufft_rt =
       cufftExecC2C(scoped_plan.plan(), idata, odata, direction);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecC2C failed");
 }
 
-template <typename ScopedPlanType>
-inline void exec_plan(ScopedPlanType& scoped_plan, cufftDoubleComplex* idata,
-                      cufftDoubleComplex* odata, int direction) {
+inline void exec_plan(const ScopedCufftPlan& scoped_plan,
+                      cufftDoubleComplex* idata, cufftDoubleComplex* odata,
+                      int direction) {
   cufftResult cufft_rt =
       cufftExecZ2Z(scoped_plan.plan(), idata, odata, direction);
   KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftExecZ2Z failed");
--- a/fft/src/KokkosFFT_Cuda_transform.hpp
+++ b/fft/src/KokkosFFT_Cuda_transform.hpp
@@ -6,12 +6,10 @@
 #define KOKKOSFFT_CUDA_TRANSFORM_HPP
 
 #include <cufft.h>
-#include "KokkosFFT_asserts.hpp"
+#include "KokkosFFT_Cuda_types.hpp"
 
 namespace KokkosFFT {
 namespace Impl {
-
-struct ScopedCufftPlan;
 
 inline void exec_plan(const ScopedCufftPlan& scoped_plan, cufftReal* idata,
                       cufftComplex* odata, int /*direction*/) {
--- a/fft/src/KokkosFFT_Cuda_transform.hpp
+++ b/fft/src/KokkosFFT_Cuda_transform.hpp
@@ -6,6 +6,7 @@
 #define KOKKOSFFT_CUDA_TRANSFORM_HPP
 
 #include <cufft.h>
+#include "KokkosFFT_asserts.hpp"
 #include "KokkosFFT_Cuda_types.hpp"
 
 namespace KokkosFFT {
--- a/fft/src/KokkosFFT_Cuda_types.hpp
+++ b/fft/src/KokkosFFT_Cuda_types.hpp
@@ -24,7 +24,6 @@ namespace Impl {
 using FFTDirectionType = int;
 
 /// \brief A class that wraps cufft for RAII
-template <typename ExecutionSpace, typename T1, typename T2>
 struct ScopedCufftPlanType {
   cufftHandle m_plan;
 
@@ -129,7 +128,7 @@ struct transform_type<ExecutionSpace, Kokkos::complex<T1>,
 template <typename ExecutionSpace, typename T1, typename T2>
 struct FFTPlanType {
   using fftw_plan_type  = ScopedFFTWPlanType<ExecutionSpace, T1, T2>;
-  using cufft_plan_type = ScopedCufftPlanType<ExecutionSpace, T1, T2>;
+  using cufft_plan_type = ScopedCufftPlanType;
   using type = std::conditional_t<std::is_same_v<ExecutionSpace, Kokkos::Cuda>,
                                   cufft_plan_type, fftw_plan_type>;
 };
@@ -192,7 +191,7 @@ struct transform_type<ExecutionSpace, Kokkos::complex<T1>,
 
 template <typename ExecutionSpace, typename T1, typename T2>
 struct FFTPlanType {
-  using type = ScopedCufftPlanType<ExecutionSpace, T1, T2>;
+  using type = ScopedCufftPlanType;
 };
 
 template <typename ExecutionSpace>
--- a/fft/src/KokkosFFT_Cuda_types.hpp
+++ b/fft/src/KokkosFFT_Cuda_types.hpp
@@ -67,7 +67,7 @@ struct ScopedCufftPlan {
   ScopedCufftPlan(ScopedCufftPlan &&)                 = delete;
 
   cufftHandle plan() const noexcept { return m_plan; }
-  void commit(const Kokkos::Cuda &exec_space) {
+  void commit(const Kokkos::Cuda &exec_space) const {
     cufftResult cufft_rt = cufftSetStream(m_plan, exec_space.cuda_stream());
     KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftSetStream failed");
   }
--- a/fft/src/KokkosFFT_Cuda_types.hpp
+++ b/fft/src/KokkosFFT_Cuda_types.hpp
@@ -5,7 +5,6 @@
 #ifndef KOKKOSFFT_CUDA_TYPES_HPP
 #define KOKKOSFFT_CUDA_TYPES_HPP
 
-#include <iostream>
 #include <cufft.h>
 #include <Kokkos_Abort.hpp>
 #include "KokkosFFT_common_types.hpp"
@@ -56,7 +55,10 @@ struct ScopedCufftPlan {
     KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftPlanMany failed");
   }
 
-  ~ScopedCufftPlan() noexcept { cleanup(); }
+  ~ScopedCufftPlan() noexcept {
+    cufftResult cufft_rt = cufftDestroy(m_plan);
+    if (cufft_rt != CUFFT_SUCCESS) Kokkos::abort("cufftDestroy failed");
+  }
 
   ScopedCufftPlan()                                   = delete;
   ScopedCufftPlan(const ScopedCufftPlan &)            = delete;
@@ -66,21 +68,8 @@ struct ScopedCufftPlan {
 
   cufftHandle plan() const noexcept { return m_plan; }
   void commit(const Kokkos::Cuda &exec_space) {
-    cudaStream_t stream = exec_space.cuda_stream();
-    try {
-      cufftResult cufft_rt = cufftSetStream(m_plan, stream);
-      KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftSetStream failed");
-    } catch (const std::runtime_error &e) {
-      std::cerr << e.what() << std::endl;
-      cleanup();
-      throw;
-    }
-  }
-
- private:
-  void cleanup() {
-    cufftResult cufft_rt = cufftDestroy(m_plan);
-    if (cufft_rt != CUFFT_SUCCESS) Kokkos::abort("cufftDestroy failed");
+    cufftResult cufft_rt = cufftSetStream(m_plan, exec_space.cuda_stream());
+    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftSetStream failed");
   }
 };
 
--- a/fft/src/KokkosFFT_Cuda_types.hpp
+++ b/fft/src/KokkosFFT_Cuda_types.hpp
@@ -6,7 +6,9 @@
 #define KOKKOSFFT_CUDA_TYPES_HPP
 
 #include <cufft.h>
+#include <Kokkos_Abort.hpp>
 #include "KokkosFFT_common_types.hpp"
+#include "KokkosFFT_asserts.hpp"
 
 #if defined(ENABLE_HOST_AND_DEVICE)
 #include "KokkosFFT_FFTW_Types.hpp"
@@ -25,10 +27,18 @@ using FFTDirectionType = int;
 
 /// \brief A class that wraps cufft for RAII
 struct ScopedCufftPlanType {
+ private:
   cufftHandle m_plan;
 
-  ScopedCufftPlanType() { cufftCreate(&m_plan); }
-  ~ScopedCufftPlanType() { cufftDestroy(m_plan); }
+ public:
+  ScopedCufftPlanType() {
+    cufftResult cufft_rt = cufftCreate(&m_plan);
+    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftCreate failed");
+  }
+  ~ScopedCufftPlanType() {
+    cufftResult cufft_rt = cufftDestroy(m_plan);
+    if (cufft_rt != CUFFT_SUCCESS) Kokkos::abort("cufftDestroy failed");
+  }
 
   cufftHandle &plan() { return m_plan; }
 };
--- a/fft/src/KokkosFFT_Cuda_types.hpp
+++ b/fft/src/KokkosFFT_Cuda_types.hpp
@@ -26,21 +26,62 @@ namespace Impl {
 using FFTDirectionType = int;
 
 /// \brief A class that wraps cufft for RAII
-struct ScopedCufftPlanType {
+template <typename ExecutionSpace>
+struct ScopedCufftPlan {
  private:
   cufftHandle m_plan;
 
  public:
-  ScopedCufftPlanType() {
-    cufftResult cufft_rt = cufftCreate(&m_plan);
-    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftCreate failed");
+  ScopedCufftPlan(const ExecutionSpace &exec_space, int nx, cufftType type,
+                  int batch) {
+    cufftResult cufft_rt = cufftPlan1d(&m_plan, nx, type, batch);
+    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftPlan1d failed");
+    set_stream(exec_space);
   }
-  ~ScopedCufftPlanType() {
+
+  ScopedCufftPlan(const ExecutionSpace &exec_space, int nx, int ny,
+                  cufftType type) {
+    cufftResult cufft_rt = cufftPlan2d(&m_plan, nx, ny, type);
+    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftPlan2d failed");
+    set_stream(exec_space);
+  }
+
+  ScopedCufftPlan(const ExecutionSpace &exec_space, int nx, int ny, int nz,
+                  cufftType type) {
+    cufftResult cufft_rt = cufftPlan3d(&m_plan, nx, ny, nz, type);
+    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftPlan3d failed");
+    set_stream(exec_space);
+  }
+
+  ScopedCufftPlan(const ExecutionSpace &exec_space, int rank, int *n,
+                  int *inembed, int istride, int idist, int *onembed,
+                  int ostride, int odist, cufftType type, int batch) {
+    cufftResult cufft_rt =
+        cufftPlanMany(&m_plan, rank, n, inembed, istride, idist, onembed,
+                      ostride, odist, type, batch);
+    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftPlanMany failed");
+    set_stream(exec_space);
+  }
+
+  ~ScopedCufftPlan() noexcept {
     cufftResult cufft_rt = cufftDestroy(m_plan);
     if (cufft_rt != CUFFT_SUCCESS) Kokkos::abort("cufftDestroy failed");
   }
 
+  ScopedCufftPlan()                                   = delete;
+  ScopedCufftPlan(const ScopedCufftPlan &)            = delete;
+  ScopedCufftPlan &operator=(const ScopedCufftPlan &) = delete;
+  ScopedCufftPlan &operator=(ScopedCufftPlan &&)      = delete;
+  ScopedCufftPlan(ScopedCufftPlan &&)                 = delete;
+
   cufftHandle &plan() { return m_plan; }
+
+ private:
+  void set_stream(const ExecutionSpace &exec_space) {
+    cudaStream_t stream  = exec_space.cuda_stream();
+    cufftResult cufft_rt = cufftSetStream(m_plan, stream);
+    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftSetStream failed");
+  }
 };
 
 #if defined(ENABLE_HOST_AND_DEVICE)
@@ -137,8 +178,8 @@ struct transform_type<ExecutionSpace, Kokkos::complex<T1>,
 
 template <typename ExecutionSpace, typename T1, typename T2>
 struct FFTPlanType {
-  using fftw_plan_type  = ScopedFFTWPlanType<ExecutionSpace, T1, T2>;
-  using cufft_plan_type = ScopedCufftPlanType;
+  using fftw_plan_type  = ScopedFFTWPlan<ExecutionSpace, T1, T2>;
+  using cufft_plan_type = ScopedCufftPlan<ExecutionSpace>;
   using type = std::conditional_t<std::is_same_v<ExecutionSpace, Kokkos::Cuda>,
                                   cufft_plan_type, fftw_plan_type>;
 };
@@ -201,7 +242,7 @@ struct transform_type<ExecutionSpace, Kokkos::complex<T1>,
 
 template <typename ExecutionSpace, typename T1, typename T2>
 struct FFTPlanType {
-  using type = ScopedCufftPlanType;
+  using type = ScopedCufftPlan<ExecutionSpace>;
 };
 
 template <typename ExecutionSpace>
--- a/fft/src/KokkosFFT_Cuda_types.hpp
+++ b/fft/src/KokkosFFT_Cuda_types.hpp
@@ -73,7 +73,6 @@ struct ScopedCufftPlan {
   ScopedCufftPlan &operator=(ScopedCufftPlan &&)      = delete;
   ScopedCufftPlan(ScopedCufftPlan &&)                 = delete;
 
-  // cufftHandle &plan() { return m_plan; }
   cufftHandle plan() const noexcept { return m_plan; }
 
  private:
--- a/fft/src/KokkosFFT_Cuda_types.hpp
+++ b/fft/src/KokkosFFT_Cuda_types.hpp
@@ -8,6 +8,10 @@
 #include <cufft.h>
 #include "KokkosFFT_common_types.hpp"
 
+#if defined(ENABLE_HOST_AND_DEVICE)
+#include "KokkosFFT_FFTW_Types.hpp"
+#endif
+
 // Check the size of complex type
 static_assert(sizeof(cufftComplex) == sizeof(Kokkos::complex<float>));
 static_assert(alignof(cufftComplex) <= alignof(Kokkos::complex<float>));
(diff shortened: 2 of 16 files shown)


tpadioleau:
I would go a step further and make these RAII classes unique-owners. That implies to disallow shallow copy either by simply deleting copy constructor and assignment or make them allocate and copy a new plan if backends offer such a copy function. We can keep the move semantics but we have to keep track if the current object must call `cufftDestroy` in the destructor. If we delete both the copy and move semantics that gives:

```cpp
class ScopedCufftPlanType {
 private:
  cufftHandle m_plan;

 public:
  ScopedCufftPlanType() {
    cufftResult cufft_rt = cufftCreate(&m_plan);
    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftCreate failed");
  }

  ScopedCufftPlanType(ScopedCufftPlanType const& rhs) = delete;

  ScopedCufftPlanType(ScopedCufftPlanType&& rhs) = delete;

  ~ScopedCufftPlanType() noexcept {
    cufftResult cufft_rt = cufftDestroy(m_plan);
    if (cufft_rt != CUFFT_SUCCESS) Kokkos::abort("cufftDestroy failed");
  }

  ScopedCufftPlanType& operator=(ScopedCufftPlanType const& rhs) = delete;

  ScopedCufftPlanType& operator=(ScopedCufftPlanType&& rhs) = delete;

  cufftHandle plan() const noexcept { return m_plan; }
};
```

The `ScopedCufftPlanType::plan` can return the backend handle by copy for CUDA, I don't know for the other backends.

I notice that we could also improve the error messages, I see that `cufftCreate` can return different error types https://docs.nvidia.com/cuda/cufft/index.html?highlight=cufftMakePlan1d#c.cufftCreate.

yasahi-hpc:
> I would go a step further and make these RAII classes unique-owners. That implies to disallow shallow copy either by simply deleting copy constructor and assignment or make them allocate and copy a new plan if backends offer such a copy function. We can keep the move semantics but we have to keep track if the current object must call `cufftDestroy` in the destructor. 

I would prefer to disallow both copy and move, because it may be unsafe for some backends. 

> If we delete both the copy and move semantics that gives:
> 
> ```c++
> class ScopedCufftPlanType {
>  private:
>   cufftHandle m_plan;
> 
>  public:
>   ScopedCufftPlanType() {
>     cufftResult cufft_rt = cufftCreate(&m_plan);
>     KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftCreate failed");
>   }
> 
>   ScopedCufftPlanType(ScopedCufftPlanType const& rhs) = delete;
> 
>   ScopedCufftPlanType(ScopedCufftPlanType&& rhs) = delete;
> 
>   ~ScopedCufftPlanType() noexcept {
>     cufftResult cufft_rt = cufftDestroy(m_plan);
>     if (cufft_rt != CUFFT_SUCCESS) Kokkos::abort("cufftDestroy failed");
>   }
> 
>   ScopedCufftPlanType& operator=(ScopedCufftPlanType const& rhs) = delete;
> 
>   ScopedCufftPlanType& operator=(ScopedCufftPlanType&& rhs) = delete;
> 
>   cufftHandle plan() const noexcept { return m_plan; }
> };
> ```
> 
> The `ScopedCufftPlanType::plan` can return the backend handle by copy for CUDA, I don't know for the other backends.

Does it return a shallow copy of the `cufftHandle`? I could not find the details of copy constructor of `cufftHandle`. That should work for oneMKL backend, but I am not sure about the behavior of other backends

> I notice that we could also improve the error messages, I see that `cufftCreate` can return different error types https://docs.nvidia.com/cuda/cufft/index.html?highlight=cufftMakePlan1d#c.cufftCreate.

That is true. We can improve the error messages

tpadioleau:
> I would prefer to disallow both copy and move, because it may be unsafe for some backends.

Ok for me. I don't think there should be any problem with implementing the move semantics later. It requires us to only pass these objects by reference.

> Does it return a shallow copy of the `cufftHandle`? I could not find the details of copy constructor of `cufftHandle`. That should work for oneMKL backend, but I am not sure about the behavior of other backends

I mean if we do

```cpp
ScopedCufftPlanType plan1;
ScopedCufftPlanType plan2;
plan2 = plan1;
```
This compiles fine and `plan2` will get the same `cufftHandle` than `plan1`. The `plan2` will destroy correctly the handle but then `plan1` will likely fail to do so.

yasahi-hpc:
> > I would prefer to disallow both copy and move, because it may be unsafe for some backends.
> 
> Ok for me. I don't think there should be any problem with implementing the move semantics later. It requires us to only pass these objects by reference.
> 
> > Does it return a shallow copy of the `cufftHandle`? I could not find the details of copy constructor of `cufftHandle`. That should work for oneMKL backend, but I am not sure about the behavior of other backends
> 
> I mean if we do
> 
> ```c++
> ScopedCufftPlanType plan1;
> ScopedCufftPlanType plan2;
> plan2 = plan1;
> ```
> 
> This compiles fine and `plan2` will get the same `cufftHandle` than `plan1`. The `plan2` will destroy correctly the handle but then `plan1` will likely fail to do so.

That is right. This does not work in the current implementation. Yes, for safety, it is better to disallow copy

fft/src/KokkosFFT_Cuda_types.hpp
```diff
@@ -15,27 +21,70 @@ static_assert(alignof(cufftComplex) <= alignof(Kokkos::complex<float>));
...
-static_assert(alignof(fftw_complex) <= alignof(Kokkos::complex<double>));
-#endif
-
 namespace KokkosFFT {
 namespace Impl {
 using FFTDirectionType = int;
 
-// Unused
+/// \brief A class that wraps cufft for RAII
 template <typename ExecutionSpace>
```
tpadioleau, line 29: Does it support other execution spaces than `Kokkos::HIP` ? If not I would remove the template argument

fft/src/KokkosFFT_Cuda_transform.hpp
```diff
@@ -10,45 +10,47 @@
 
 namespace KokkosFFT {
 namespace Impl {
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftReal* idata, cufftComplex* odata,
-                      int /*direction*/, Args...) {
-  cufftResult cufft_rt = cufftExecR2C(plan, idata, odata);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftReal* idata,
+                      cufftComplex* odata, int /*direction*/) {
```
tpadioleau, line 15: Do these functions work with a `ScopedPlanType` different than a `ScopedCufftPlan` ? If not I would remove the template argument.

fft/src/KokkosFFT_Cuda_transform.hpp
```diff
@@ -10,45 +10,47 @@
 
 namespace KokkosFFT {
 namespace Impl {
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftReal* idata, cufftComplex* odata,
-                      int /*direction*/, Args...) {
-  cufftResult cufft_rt = cufftExecR2C(plan, idata, odata);
+template <typename ScopedPlanType>
+inline void exec_plan(ScopedPlanType& scoped_plan, cufftReal* idata,
```
tpadioleau, line 14: Can you pass the `scoped_plan` by `const&` ?

fft/src/KokkosFFT_Cuda_types.hpp
```diff
@@ -15,27 +21,70 @@ static_assert(alignof(cufftComplex) <= alignof(Kokkos::complex<float>));
...
+    if (cufft_rt != CUFFT_SUCCESS) Kokkos::abort("cufftDestroy failed");
+  }
+
+  ScopedCufftPlan()                                   = delete;
+  ScopedCufftPlan(const ScopedCufftPlan &)            = delete;
+  ScopedCufftPlan &operator=(const ScopedCufftPlan &) = delete;
+  ScopedCufftPlan &operator=(ScopedCufftPlan &&)      = delete;
+  ScopedCufftPlan(ScopedCufftPlan &&)                 = delete;
+
+  cufftHandle &plan() { return m_plan; }
```
tpadioleau, line 77:
```suggestion
  cufftHandle plan() const noexcept { return m_plan; }
```

fft/src/KokkosFFT_FFTW_Types.hpp
```diff
@@ -0,0 +1,154 @@
...
+      m_plan =
+          fftwf_plan_many_dft(rank, n, howmany, in, inembed, istride, idist,
+                              out, onembed, ostride, odist, sign, flags);
+    } else if constexpr (type == KokkosFFT::Impl::FFTWTransformType::Z2Z) {
+      m_plan = fftw_plan_many_dft(rank, n, howmany, in, inembed, istride, idist,
+                                  out, onembed, ostride, odist, sign, flags);
+    }
+    m_is_created = true;
+  }
+  ~ScopedFFTWPlan() noexcept {
```
tpadioleau, line 108:
```suggestion
  }

  ~ScopedFFTWPlan() noexcept {
```

fft/src/KokkosFFT_FFTW_Types.hpp
```diff
@@ -0,0 +1,154 @@
...
+    cleanup_threads();
+    if constexpr (std::is_same_v<plan_type, fftwf_plan>) {
+      if (m_is_created) fftwf_destroy_plan(m_plan);
+    } else {
+      if (m_is_created) fftw_destroy_plan(m_plan);
+    }
```
tpadioleau, line 114: In which situation do we end up with `m_is_created == false` ?

fft/src/KokkosFFT_FFTW_Types.hpp
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  ScopedFFTWPlan()                                  = delete;
+  ScopedFFTWPlan(const ScopedFFTWPlan &)            = delete;
+  ScopedFFTWPlan &operator=(const ScopedFFTWPlan &) = delete;
+  ScopedFFTWPlan &operator=(ScopedFFTWPlan &&)      = delete;
+  ScopedFFTWPlan(ScopedFFTWPlan &&)                 = delete;
+
+  const plan_type &plan() const { return m_plan; }
```
tpadioleau, line 123:
```suggestion
  plan_type plan() const noexcept { return m_plan; }
```

fft/src/KokkosFFT_HIP_types.hpp
```diff
@@ -15,27 +21,70 @@ static_assert(alignof(hipfftComplex) <= alignof(Kokkos::complex<float>));
...
+    if (hipfft_rt != HIPFFT_SUCCESS) Kokkos::abort("hipfftDestroy failed");
+  }
+
+  ScopedHIPfftPlan()                                    = delete;
+  ScopedHIPfftPlan(const ScopedHIPfftPlan &)            = delete;
+  ScopedHIPfftPlan &operator=(const ScopedHIPfftPlan &) = delete;
+  ScopedHIPfftPlan &operator=(ScopedHIPfftPlan &&)      = delete;
+  ScopedHIPfftPlan(ScopedHIPfftPlan &&)                 = delete;
+
+  hipfftHandle &plan() { return m_plan; }
```
tpadioleau, line 77:
```suggestion
  hipfftHandle plan() const noexcept { return m_plan; }
```

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -272,16 +259,13 @@ class Plan {
...
-  }
 
-  ~Plan() {
-    KokkosFFT::Impl::destroy_plan_and_info<ExecutionSpace, fft_plan_type,
-                                           fft_info_type>(m_plan, m_info);
+    m_fft_size = KokkosFFT::Impl::create_plan(exec_space, m_plan, in, out,
+                                              direction, axes, s, m_is_inplace);
   }
 
+  ~Plan() {}
```
tpadioleau, line 267:
```suggestion
  ~Plan() noexcept = default;
```

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +23,67 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+      Kokkos::View<Kokkos::complex<floating_point_type> *, ExecutionSpace>;
+
+  bool m_is_plan_created = false;
+  bool m_is_info_created = false;
+
+  //! Internal work buffer
+  BufferViewType m_buffer;
+
+ public:
+  ScopedRocfftPlan() {}
```
tpadioleau, line 57:
```suggestion
  ScopedRocfftPlan() = default;
```

Is it required though ?

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +23,67 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+  ScopedRocfftPlan &operator=(ScopedRocfftPlan &&)      = delete;
+  ScopedRocfftPlan(ScopedRocfftPlan &&)                 = delete;
+
+  void set_is_plan_created() { m_is_plan_created = true; }
+  void set_is_info_created() { m_is_info_created = true; }
+
+  void allocate_work_buffer(std::size_t workbuffersize) {
+    m_buffer = BufferViewType("workbuffer", workbuffersize);
+  }
+  rocfft_plan &plan() { return m_plan; }
```
tpadioleau, line 82:
```suggestion
  rocfft_plan plan() const noexcept { return m_plan; }
```

fft/src/KokkosFFT_ROCM_plans.hpp
```diff
@@ -129,8 +127,13 @@ auto create_plan(const ExecutionSpace& exec_space,
...
 
   // Create the description
-  rocfft_plan_description description;
-  rocfft_status status = rocfft_plan_description_create(&description);
+  std::unique_ptr<rocfft_plan_description,
+                  std::function<void(rocfft_plan_description*)>> const
+      description(new rocfft_plan_description,
+                  [](rocfft_plan_description* desc) {
+                    rocfft_plan_description_destroy(*desc);
+                  });
```
tpadioleau, line 135:
```suggestion
                  [](rocfft_plan_description* desc) {
                    rocfft_plan_description_destroy(*desc);
                    delete desc;
                  });
```

fft/src/KokkosFFT_ROCM_plans.hpp
```diff
@@ -129,8 +127,13 @@ auto create_plan(const ExecutionSpace& exec_space,
...
+  rocfft_status status = rocfft_plan_description_create(&(*description));
```
tpadioleau, line 136: I think we can improve the situation by also wrapping `rocfft_plan_description`.

In case `rocfft_plan_description_create` fails, the next `KOKKOSFFT_THROW_IF` will throw, calling the destructor of `description` thus calling `rocfft_plan_description_destroy` even though the `rocfft_plan_description` object has not been initialised.

fft/src/KokkosFFT_ROCM_plans.hpp
```diff
@@ -159,56 +162,46 @@ auto create_plan(const ExecutionSpace& exec_space,
...
   // Create a plan
   plan   = std::make_unique<PlanType>();
-  status = rocfft_plan_create(&(*plan), place, fft_direction, precision,
+  status = rocfft_plan_create(&(plan->plan()), place, fft_direction, precision,
                               reversed_fft_extents.size(),  // Dimension
                               reversed_fft_extents.data(),  // Lengths
-                              howmany,     // Number of transforms
-                              description  // Description
+                              howmany,      // Number of transforms
+                              *description  // Description
```
tpadioleau, line 169: Why not making this part of the constructor of `ScopedRocfftPlan` ?

fft/src/KokkosFFT_FFTW_Types.hpp
yasahi-hpc: I may expect that `fftw_plan_many_dft` may fail if we request too much memory

fft/src/KokkosFFT_Cuda_types.hpp
yasahi-hpc: I have fixed accordingly. Now using `Kokkos::HIP` explicitly

fft/src/KokkosFFT_Cuda_transform.hpp
yasahi-hpc: Now using `ScopedCufftPlan` explicitly

fft/src/KokkosFFT_Cuda_transform.hpp
yasahi-hpc: Modified accordingly

fft/src/KokkosFFT_Cuda_types.hpp
yasahi-hpc: Sure

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: `ScopedRocfftPlan` was refactored and this constructor was removed

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: Sure

fft/src/KokkosFFT_ROCM_plans.hpp
yasahi-hpc: I introduced a wrapper for description as well for RAII. This part has been removed

fft/src/KokkosFFT_ROCM_plans.hpp
yasahi-hpc: I have introduced a wrapper accordingly

fft/src/KokkosFFT_ROCM_plans.hpp
yasahi-hpc: I moved the logic into the constructor

yasahi-hpc (commented):
@tpadioleau 
Thank you for reviews. I have fixed accordingly

fft/src/KokkosFFT_FFTW_Types.hpp
tpadioleau: That is a possibility but I am not sure to understand how we capture this, I don't any error status variable in this snippet ?

fft/src/KokkosFFT_Cuda_transform.hpp
```diff
@@ -10,45 +10,47 @@
 
 namespace KokkosFFT {
 namespace Impl {
-template <typename... Args>
-inline void exec_plan(cufftHandle& plan, cufftReal* idata, cufftComplex* odata,
-                      int /*direction*/, Args...) {
-  cufftResult cufft_rt = cufftExecR2C(plan, idata, odata);
+
+struct ScopedCufftPlan;
```
tpadioleau, line 14: I feel we should include `KokkosFFT_Cuda_type.hpp` instead because the code below requires the complete definition of the `ScopedCufftPlan` ?

fft/src/KokkosFFT_Host_transform.hpp
```diff
@@ -9,40 +9,41 @@
 
 namespace KokkosFFT {
 namespace Impl {
-template <typename PlanType, typename... Args>
-void exec_plan(PlanType& plan, float* idata, fftwf_complex* odata,
-               int /*direction*/, Args...) {
-  fftwf_execute_dft_r2c(plan, idata, odata);
+
+template <typename ScopedPlanType>
```
tpadioleau, line 13: Can we also replace this template parameter by`ScopedFFTWPlan` ?

fft/src/KokkosFFT_ROCM_transform.hpp
```diff
@@ -11,60 +11,65 @@
 
 namespace KokkosFFT {
 namespace Impl {
-inline void exec_plan(rocfft_plan& plan, float* idata,
-                      std::complex<float>* odata, int /*direction*/,
-                      const rocfft_execution_info& execution_info) {
+template <typename ScopedPlanType>
```
tpadioleau, line 14: Same question for `ScopedRocfftPlan`

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +25,227 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+    KOKKOSFFT_THROW_IF(status != rocfft_status_success,
+                       "rocfft_plan_description_create failed");
+  }
+  ~ScopedRocfftPlanDescription() noexcept {
+    rocfft_status status = rocfft_plan_description_destroy(m_description);
+    if (status != rocfft_status_success)
+      Kokkos::abort("rocfft_plan_description_destroy failed");
+  }
+
+  rocfft_plan_description &description() { return m_description; }
```
tpadioleau, line 58:
```suggestion
  rocfft_plan_description description() const noexcept { return m_description; }
```

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +25,227 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+  }
+
+  ScopedRocfftPlan()                                    = delete;
+  ScopedRocfftPlan(const ScopedRocfftPlan &)            = delete;
+  ScopedRocfftPlan &operator=(const ScopedRocfftPlan &) = delete;
+  ScopedRocfftPlan &operator=(ScopedRocfftPlan &&)      = delete;
+  ScopedRocfftPlan(ScopedRocfftPlan &&)                 = delete;
+
+  rocfft_plan plan() const noexcept { return m_plan; }
+  rocfft_execution_info &execution_info() { return m_execution_info; }
```
tpadioleau, line 185:
```suggestion
  rocfft_execution_info execution_info() const noexcept { return m_execution_info; }
```

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +25,227 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+
+    m_is_info_created = true;
+
+    // set stream
+    // NOTE: The stream must be of type hipStream_t.
+    // It is an error to pass the address of a hipStream_t object.
+    hipStream_t stream = exec_space.hip_stream();
+    status = rocfft_execution_info_set_stream(m_execution_info, stream);
+    KOKKOSFFT_THROW_IF(status != rocfft_status_success,
+                       "rocfft_execution_info_set_stream failed");
```
tpadioleau, line 149: I would also wrap `rocfft_execution_info`. If `rocfft_execution_info_set_stream` fails, `m_execution_info` will never be freed.

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +25,227 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+
+    // Create a plan
+    status = rocfft_plan_create(&m_plan, place, fft_direction, m_precision,
+                                reversed_fft_extents.size(),  // Dimension
+                                reversed_fft_extents.data(),  // Lengths
+                                howmany,  // Number of transforms
+                                scoped_description.description()  // Description
+    );
+    KOKKOSFFT_THROW_IF(status != rocfft_status_success,
+                       "rocfft_plan_create failed");
```
tpadioleau, line 132: I think we should try to make the last thing to call in the constructor, if there is any exception throw after the plan creation, the destructor will never be called and the plan will never to freed.

fft/src/KokkosFFT_Cuda_transform.hpp
yasahi-hpc: Sure

fft/src/KokkosFFT_Host_transform.hpp
yasahi-hpc: It is not that easy.
`ScopedFFTWPlan <ExecSpace, T1, T2>` is a class template, where `ExecSpace ` can be either `Kokkos::Serial`, `Kokkos::Threads`, or `Kokkos::OpenMP`
To make the matter worse, these `T1` and `T2` are not the types of `idata` and `odata`. 
Thus, I think it is better to use template here

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: Sure. 
If there is a performance issue due to returning by value, I may change this

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: Sure

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: That is true. I will do that

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: Could you please detail this point?
If there is an exception after plan creation, `m_plan` is supposed to be destroyed in destructor of `ScopedRocfftPlan`

fft/src/KokkosFFT_FFTW_Types.hpp
yasahi-hpc: If I understand correctly, the code will abort if something gets wrong with `fftw`. So may be this check is not useful

yasahi-hpc:
@tpadioleau 
Thank you again for your review. 
After some consideration, I have introduced `try catch` inside a constructor and cleanup manually if something gets wrong. If an exception is thrown in constructor, we cannot rely on the destructor, so we need to cleanup by ourselves.

fft/src/KokkosFFT_Cuda_transform.hpp
```diff
@@ -6,49 +6,49 @@
 #define KOKKOSFFT_CUDA_TRANSFORM_HPP
 
 #include <cufft.h>
-#include "KokkosFFT_asserts.hpp"
+#include "KokkosFFT_Cuda_types.hpp"
```
tpadioleau, line 9:
```suggestion
#include "KokkosFFT_asserts.hpp"
#include "KokkosFFT_Cuda_types.hpp"
```

I don't we want to promise that `KokkosFFT_Cuda_types.hpp` brings `KokkosFFT_asserts.hpp`

fft/src/KokkosFFT_Cuda_types.hpp
```diff
@@ -15,27 +22,95 @@ static_assert(alignof(cufftComplex) <= alignof(Kokkos::complex<float>));
...
+                  int batch) {
+    try {
+      cufftResult cufft_rt = cufftPlan1d(&m_plan, nx, type, batch);
+      KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftPlan1d failed");
+      set_stream(exec_space);
+    } catch (const std::runtime_error &e) {
+      std::cerr << e.what() << std::endl;
+      cleanup();
+      throw;
+    }
```
tpadioleau, line 45:
```suggestion
    cufftResult cufft_rt = cufftPlan1d(&m_plan, nx, type, batch);
    KOKKOSFFT_THROW_IF(cufft_rt != CUFFT_SUCCESS, "cufftPlan1d failed");
    try {
      set_stream(exec_space);
    } catch (const std::runtime_error &e) {
      std::cerr << e.what() << std::endl;
      cleanup();
      throw;
    }
```

We don't want to cleanup if `cufftPlan1d` fails.

This should work indeed but I think this shows that we try to do too many things in the constructor thus making it complicated to read and reason about. I think we should separate the plan creation from the stream installation, like they do in SYCL they have a `commit` function that takes the `sycl::queue` right ?

fft/src/KokkosFFT_Host_transform.hpp
tpadioleau: We could still use templates but for the plan:
```
template <typename ExecutionSpace, typename T1, typename T2>
void exec_plan(ScopedFFTWPlan<ExecutionSpace, T1, T2> const& scoped_plan, float* idata, fftwf_complex* odata,
               int /*direction*/) {
  fftwf_execute_dft_r2c(scoped_plan.plan(), idata, odata);
}
```

This way we would know better what input is expected don't you think ?

fft/src/KokkosFFT_Host_transform.hpp
```diff
@@ -9,40 +9,41 @@
...
+void exec_plan(ScopedPlanType& scoped_plan, float* idata, fftwf_complex* odata,
```
tpadioleau, line 14:
```suggestion
void exec_plan(ScopedPlanType const& scoped_plan, float* idata, fftwf_complex* odata,
```

Same for others

fft/src/KokkosFFT_ROCM_transform.hpp
```diff
@@ -7,64 +7,69 @@
...
 #include <rocfft/rocfft.h>
-#include "KokkosFFT_asserts.hpp"
+#include "KokkosFFT_ROCM_types.hpp"
 
 namespace KokkosFFT {
 namespace Impl {
-inline void exec_plan(rocfft_plan& plan, float* idata,
-                      std::complex<float>* odata, int /*direction*/,
-                      const rocfft_execution_info& execution_info) {
+inline void exec_plan(const ScopedRocfftPlan<float>& scoped_plan, float* idata,
```
tpadioleau, line 14:
```suggestion
inline void exec_plan(const ScopedRocfftPlan<float> const& scoped_plan, float* idata,
```

Same for others

tpadioleau (commented):
I am not convinced this is the right approach. We can further discuss it in january if you want.

fft/src/KokkosFFT_ROCM_types.hpp
tpadioleau: Maybe there was confusion, by plan creation I meant the creation of `m_plan`.

fft/src/KokkosFFT_Cuda_types.hpp
yasahi-hpc: > We don't want to cleanup if cufftPlan1d fails.

Sure

> This should work indeed but I think this shows that we try to do too many things in the constructor thus making it complicated to read and reason about. I think we should separate the plan creation from the stream installation, like they do in SYCL they have a commit function that takes the sycl::queue right ?

The problem is that the lifetime of scoped plan instance is identical to the lifetime of Plan class, since we need to keep it for potential execution of the plan.

https://github.com/kokkos/kokkos-fft/blob/95bd3b317201087e73cc4bf9786158ef22087383/fft/src/KokkosFFT_Plans.hpp#L111-L112

Thus, even if scoped plan instance is constructed, its destructor will not be called if the destructor of Plan class is not called (this is the case if the Plan class construction is disrupted by exception). 
In principle, we have the situation like this
https://godbolt.org/z/veMhKsfjh

If `commit` function throws, we will have the same issue. Thus, I think we need to cleanup by ourselves.

fft/src/KokkosFFT_Cuda_transform.hpp
yasahi-hpc: Sure

fft/src/KokkosFFT_Cuda_types.hpp
yasahi-hpc: If we remove the `try catch`, I got some errors which lead to memory leaks.
https://godbolt.org/z/8KrfKh81a

tpadioleau: This is actually expected, if there is no try/catch block, there is no unwinding and the program can terminate immediately. It is the responsibility of the user of Kokkos-fft to put the code in a try/catch block.

yasahi-hpc: So your point is that

1. w/o try catch
If the `scoped plan` construction fails, the program will be terminated anyway, thus we are good

2. w/ try catch
If the `scoped plan` construction fails, the scoped plan will be destroyed appropriately. 

Then, I am fine to move the logics in the constructor to `commit` method

tpadioleau: Yes I think that's it. But in my opinion we don't even have to think about whether the constructor is called inside a try/catch block. We have to ensure that:
- if the constructor does not complete, all dynamic allocations are freed without relying on the destructor
- if it does complete, the destructor will free all dynamic allocations.

In a C++ binding over a C object oriented library we can almost have an automatic approach. We can wrap all C opaque objects in a corresponding C++ class that will only call the `*_create` function in the constructor and the `*_destroy` function in the destructor. The rest can, and I think should, be done in extra member functions in order to keep the logic simple in the constructor.

What do you think ?

tpadioleau: For example gtkmm, a C++ binding to gtk, is almost automatically generated by a set of rules making the link between the C and C++ objects.

yasahi-hpc: I mostly agree with you.

> Yes I think that's it. But in my opinion we don't even have to think about whether the constructor is called inside a try/catch block. We have to ensure that:
> 
> * if the constructor does not complete, all dynamic allocations are freed without relying on the destructor

If plan construction fails, we do not need to destroy it because it is not created? 
Then, it can be achieved except for `rocfft`. For `rocfft`, we need a special cleanup method to destroy two different objects (`plan` and `info`).

> * if it does complete, the destructor will free all dynamic allocations.

Sure. It can be easily implemented.

Do you agree that we need a special logic for `rocfft`?

yasahi-hpc:
@tpadioleau I have fixed based on the discussion. Can I have another review?

fft/src/KokkosFFT_FFTW_Types.hpp
```diff
@@ -0,0 +1,158 @@
...
+  ScopedFFTWPlan()                                  = delete;
+  ScopedFFTWPlan(const ScopedFFTWPlan &)            = delete;
+  ScopedFFTWPlan &operator=(const ScopedFFTWPlan &) = delete;
+  ScopedFFTWPlan &operator=(ScopedFFTWPlan &&)      = delete;
+  ScopedFFTWPlan(ScopedFFTWPlan &&)                 = delete;
+
+  plan_type plan() const noexcept { return m_plan; }
+
+ private:
+  void init_threads([[maybe_unused]] const ExecutionSpace &exec_space) {
```
tpadioleau, line 124:
```suggestion
  static void init_threads([[maybe_unused]] const ExecutionSpace &exec_space) noexcept {
```

The function is independent of the state of the object.

fft/src/KokkosFFT_FFTW_Types.hpp
```diff
@@ -0,0 +1,158 @@
...
+        fftwf_plan_with_nthreads(nthreads);
+      } else {
+        fftw_init_threads();
+        fftw_plan_with_nthreads(nthreads);
+      }
+    }
+#endif
+  }
+
+  void cleanup_threads() {
```
tpadioleau, line 141:
```suggestion
  static void cleanup_threads() noexcept {
```

The function is independent of the state of the object.

fft/src/KokkosFFT_ROCM_types.hpp
tpadioleau, line 1: I think we also need a `ScopedRocfftExecutionInfo` ?

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +26,228 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+    int odist = std::accumulate(out_extents.begin(), out_extents.end(), 1,
+                                std::multiplies<>());
+
+    auto in_strides  = compute_strides<int, std::size_t>(in_extents);
+    auto out_strides = compute_strides<int, std::size_t>(out_extents);
+    auto reversed_fft_extents =
+        convert_int_type_and_reverse<int, std::size_t>(fft_extents);
+
+    // Create a plan description
+    ScopedRocfftPlanDescription scoped_description;
```
tpadioleau, line 100: I think the lifetime of the rofft description object must be longer than the plan.

tpadioleau:
Why do we need the rocfft backend already ? For performance ?

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: According to the docs, this can be destroyed after passed to the plan.

https://rocm.docs.amd.com/projects/rocFFT/en/docs-5.2.3/api.html#plan-description

yasahi-hpc:
> Why do we need the rocfft backend already ? For performance ?

For performance. I prefer `rocfft` over `hipfft`

fft/src/KokkosFFT_ROCM_types.hpp
tpadioleau: Oh good I missed that part and chatGPT was clear that it needed to be destroyed after the plan... then let's keep it that way!

yasahi-hpc:
> > For performance. I prefer `rocfft` over `hipfft`
> 
> I don't understand how we can explain that ?

Me neither.
When I measured the 2D batched FFT with a reused plan, I measured a better performance with `rocfft`.
Even if the overhead is small, `rocfft` must outperform `hipfft` since `hipfft` just wraps `rocfft`

fft/src/KokkosFFT_FFTW_Types.hpp
```diff
@@ -0,0 +1,158 @@
...
+          fftwf_plan_many_dft(rank, n, howmany, in, inembed, istride, idist,
+                              out, onembed, ostride, odist, sign, flags);
+    } else if constexpr (type == KokkosFFT::Impl::FFTWTransformType::Z2Z) {
+      m_plan = fftw_plan_many_dft(rank, n, howmany, in, inembed, istride, idist,
+                                  out, onembed, ostride, odist, sign, flags);
+    }
+  }
+
+  ~ScopedFFTWPlan() noexcept {
+    cleanup_threads();
```
tpadioleau, line 107: It looks better but I am afraid it easily leads to UB. If we take the following scenario:
```cpp
KokkosFFt::Plan plan( /* ... */ );
KokkosFFt::fft( /* ... */ );
plan.execute( /* ... */ );
``` 
My understanding is that the call to `KokkosFFt::fft` will cleanup threads in the destructor of the internal `KokkosFFT::Plan`. The execution of the second FFT transform with `plan.execute` is done after a call to `fftw_cleanup_threads`, which is not allowed by FFTW https://www.fftw.org/doc/Usage-of-Multi_002dthreaded-FFTW.html

fft/src/KokkosFFT_HIP_types.hpp
```diff
@@ -15,27 +21,59 @@ static_assert(alignof(hipfftComplex) <= alignof(Kokkos::complex<float>));
...
+  }
+
+  ScopedHIPfftPlan()                                    = delete;
+  ScopedHIPfftPlan(const ScopedHIPfftPlan &)            = delete;
+  ScopedHIPfftPlan &operator=(const ScopedHIPfftPlan &) = delete;
+  ScopedHIPfftPlan &operator=(ScopedHIPfftPlan &&)      = delete;
+  ScopedHIPfftPlan(ScopedHIPfftPlan &&)                 = delete;
+
+  hipfftHandle plan() const noexcept { return m_plan; }
+  void commit(const Kokkos::HIP &exec_space) {
```
tpadioleau, line 70:
```suggestion
  void commit(const Kokkos::HIP &exec_space) const {
```

Same

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +26,242 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
+      status   = rocfft_execution_info_set_work_buffer(
+          m_execution_info, (void *)m_buffer.data(), workbuffersize);
+      KOKKOSFFT_THROW_IF(status != rocfft_status_success,
+                         "rocfft_execution_info_set_work_buffer failed");
+    }
+  }
+};
+
+/// \brief A class that wraps rocfft for RAII
+template <typename T>
```
tpadioleau, line 112: I am not convinced this class needs to be template but this can be discussed after #218

tpadioleau (commented):
I think we are almost there, just a concern with the FFTW backend and threads cleanup

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: Sure, I will do that after this

fft/src/KokkosFFT_FFTW_Types.hpp
yasahi-hpc: I tried to fix that with an unique id of `ScopedFFTWPlan`.

yasahi-hpc: I found the current implementation is sill incorrect. We need to call `fftw_cleanup_threads` in the destructor of the final instance of `ScopedFFTWPlan`. However, there is no way to identify the final instance. 

For the moment, we need to suppress `cleanup_threads()`.
We may have two options. 

1. Introduce `KokkosFFT::initialize` and `KokkosFFT::finalize` to wrap global initialization and cleanup
2. Parallelize fftw exec manually by `Kokkos`.

tpadioleau: > I found the current implementation is sill incorrect. We need to call `fftw_cleanup_threads` in the destructor of the final instance of `ScopedFFTWPlan`. However, there is no way to identify the final instance.
> 
> For the moment, we need to suppress `cleanup_threads()`. We may have two options.
> 
> 1. Introduce `KokkosFFT::initialize` and `KokkosFFT::finalize` to wrap global initialization and cleanup
> 2. Parallelize fftw exec manually by `Kokkos`.

I think I have a slight preference for the solution 1. as it will also help to solve #217. What do you think ?

yasahi-hpc: I may have another solution with a [local static object](https://godbolt.org/z/EMbYa5Ta7)

yasahi-hpc: How about removing `fftw_cleanup_threads` from this PR and make it by solution 1 in another PR?

tpadioleau: Yes looks a good idea

yasahi-hpc: @tpadioleau May I get another review. Note it is safe to call `fftw_init_threads` multiplie times.

fft/src/KokkosFFT_FFTW_Types.hpp
```diff
@@ -0,0 +1,145 @@
+// SPDX-FileCopyrightText: (C) The kokkos-fft development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+#ifndef KOKKOSFFT_FFTW_TYPES_HPP
+#define KOKKOSFFT_FFTW_TYPES_HPP
+
+#include <mutex>
```
tpadioleau, line 8:
```suggestion
```

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -5,9 +5,18 @@
 #ifndef KOKKOSFFT_ROCM_TYPES_HPP
 #define KOKKOSFFT_ROCM_TYPES_HPP
 
+#include <numeric>
+#include <algorithm>
 #include <complex>
+#include <iostream>
```
tpadioleau, line 11:
```suggestion
```

Not used right ?

fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -17,27 +26,242 @@ static_assert(sizeof(std::complex<double>) == sizeof(Kokkos::complex<double>));
...
 
+#if !defined(ENABLE_HOST_AND_DEVICE)
 enum class FFTWTransformType { R2C, D2Z, C2R, Z2D, C2C, Z2Z };
+#endif
 
 template <typename ExecutionSpace>
 using TransformType = FFTWTransformType;
 
+/// \brief A class that wraps rocfft_plan_description for RAII
+struct ScopedRocfftPlanDescription {
```
tpadioleau, line 43: If we stick to the rules of 0 and 5, we should define the copy/move constructors/assignment operators. We can delete them as for the other clases introduced in this PR.

tpadioleau (changes requested):
Minor changes and then we are good

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: True. I have removed this

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: Sure

yasahi-hpc:
@tpadioleau Thank you.
I have fixed accordingly. 
If you feel good, I will merge this and start working on #217 and #218

tpadioleau (approved):
This is fine for me, please can you add me as a git co-author for this PR ? As no suggestion was applied I will not appear at all.

PR #212, opened 2024-12-05
Remove unnecessary cufftCreate and hipfftCreate to avoid creating plans twice

### LINKED ISSUE #211: `cufftHandle` is created twice
I suspect we create twice the CUDA plan, once with `cufftCreate` and again with `cufftPlan1d` https://github.com/kokkos/kokkos-fft/blob/4b82641937cda86d56a5137ed32ba9b4a2940026/fft/src/KokkosFFT_Cuda_plans.hpp#L39-L54

From what I understand in the cuFFT documentation, we should either
- call `cufftCreate` then `cufftMakePlan1d`, https://docs.nvidia.com/cuda/cufft/index.html?highlight=cufftMakePlan1d#cufftmakeplan1d
- call directly `cufftPlan1d`, https://docs.nvidia.com/cuda/cufft/index.html?highlight=cufftMakePlan1d#cufftplan1d

Notice that both `cufftCreate` and `cufftPlan1d` take the plan by pointer whereas `cufftMakePlan1d` takes it by copy.

yasahi-hpc:
Fixes #211 

In addition, adding check functions after binding streams

### DIFF
--- a/fft/src/KokkosFFT_Cuda_plans.hpp
+++ b/fft/src/KokkosFFT_Cuda_plans.hpp
@@ -35,13 +35,6 @@ auto create_plan(const ExecutionSpace& exec_space,
   using in_value_type  = typename InViewType::non_const_value_type;
   using out_value_type = typename OutViewType::non_const_value_type;
 
-  plan                 = std::make_unique<PlanType>();
-  cufftResult cufft_rt = cufftCreate(&(*plan));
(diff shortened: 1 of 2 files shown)


tpadioleau (approved):
Yes lgtm

yasahi-hpc:
Thanks! I will merge this and update #208 on top of it

PR #219, opened 2025-01-08
fix: buffer size for rocfft execution info

### LINKED ISSUE #218: Work buffer size too big for the rocfft backend
I think we allocate too much memory in the rocfft backend. The function `rocfft_plan_get_work_buffer_size` https://rocm.docs.amd.com/projects/rocFFT/en/latest/reference/api.html#_CPPv432rocfft_plan_get_work_buffer_sizeK11rocfft_planP6size_t returns a size in bytes.

If I am not mistaken we allocate a buffer of  type `Kokkos::complex<float_type*>` in
https://github.com/kokkos/kokkos-fft/blob/95bd3b317201087e73cc4bf9786158ef22087383/fft/src/KokkosFFT_ROCM_plans.hpp#L184-L195 because `BufferViewType` is given by https://github.com/kokkos/kokkos-fft/blob/95bd3b317201087e73cc4bf9786158ef22087383/fft/src/KokkosFFT_Plans.hpp#L101-L102

yasahi-hpc:
Fixes #218 

Using a raw pointer to manage the `workbuffer`. Also we have modified the `ScopedRocfftExecutionInfo` class into a non-templated class.

### DIFF
--- a/fft/src/KokkosFFT_ROCM_types.hpp
+++ b/fft/src/KokkosFFT_ROCM_types.hpp
@@ -66,15 +66,12 @@ struct ScopedRocfftPlanDescription {
 };
 
 /// \brief A class that wraps rocfft_execution_info for RAII
-template <typename FloatingPointType>
 struct ScopedRocfftExecutionInfo {
  private:
-  using BufferViewType =
-      Kokkos::View<Kokkos::complex<FloatingPointType> *, Kokkos::HIP>;
   rocfft_execution_info m_execution_info;
 
   //! Internal work buffer
-  BufferViewType m_buffer;
+  void *m_workbuffer = nullptr;
 
  public:
   ScopedRocfftExecutionInfo() {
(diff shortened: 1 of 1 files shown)


fft/src/KokkosFFT_ROCM_types.hpp
```diff
@@ -111,9 +112,10 @@ struct ScopedRocfftExecutionInfo {
 
     // Set work buffer
     if (workbuffersize > 0) {
-      m_buffer = BufferViewType("workbuffer", workbuffersize);
-      status   = rocfft_execution_info_set_work_buffer(
-          m_execution_info, (void *)m_buffer.data(), workbuffersize);
+      hipError_t hip_status = hipMalloc(&m_workbuffer, workbuffersize);
```
tpadioleau, line 115: What do you think of using `Kokkos::kokkos_malloc` ? This would allow to trace the memory allocation in Kokkos tools.

tpadioleau (approved):
LGTM, note that we could also use a Kokkos::View with the type `std::byte`.

fft/src/KokkosFFT_ROCM_types.hpp
yasahi-hpc: That may be a good idea. I will replace this with `Kokkos::kokkos_malloc`.

yasahi-hpc: Sure
I will try that.

tpadioleau (approved):

yasahi-hpc:
@tpadioleau Thank you for the review. I will merge this

PR #229, opened 2025-02-20
Introduce cmake-format in CI

yasahi-hpc:
This PR aims at introducing CMake format check with ~~[gersemi](https://github.com/BlankSpruce/gersemi)~~ [cmake-format](https://github.com/cheshirekow/cmake_format).


JBludau:
in kokkos/kokkos we use `cmake-format` and also have an action to check it. If you want less maintenance burden you could think about copying our scripts

yasahi-hpc:
That makes sense. Maybe it is better to apply the same format as Kokkos for the consistency of the ecosystem.

.github/workflows/__cmake-format-check.yaml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - name: cmake-format lint action
+      uses: puneetmatharu/cmake-format-lint-action@2ac346e79e7ceac958bc637c1391285fb335ed7c # v1.0.5
+      with:
+        args: --config-files .cmake-format.py --in-place
+        file-regex: '(.*\.cmake$|.*\.cmake\.in$|CMakeLists.txt$)'
+    - name: check
+      run: git diff --exit-code
```
tpadioleau, line 23: Why the option `--check` is not enough ?

.github/workflows/__clang-format-check.yaml
tpadioleau, line 1: Can you add a small description about this change ?

tpadioleau (approved):
ok for me

tpadioleau:
Consider adding a small explanation on how to apply these tools for contributors ?

JBludau:
> Consider adding a small explanation on how to apply these tools for contributors ?

If we do we should do this on Kokkos, too

.github/workflows/__cmake-format-check.yaml
JBludau: It does modifications in place and we check the diff to the files in git to have printed output

yasahi-hpc: I would like to print the changes we need. 
Basically, the same as lint.

yasahi-hpc: Oh, Jacob already explained. 
Thank you for the comments

yasahi-hpc:
> Consider adding a small explanation on how to apply these tools for contributors ?

Actually, I am considering to add a developers note to the documentation in another PR. cmake-format will be explained there as well as clang-format. 

> If we do we should do this on Kokkos, too

I think it would be useful for kokkos

.github/workflows/__cmake-format-check.yaml
tpadioleau: ok thx

.github/workflows/__clang-format-check.yaml
tpadioleau: That is fine with me

tpadioleau:
> Actually, I am considering to add a developers note to the documentation in another PR. cmake-format will be explained there as well as clang-format.

Ok then it can wait

tpadioleau (approved):
Ok for me

yasahi-hpc:
@tpadioleau Thank you for approval. I will merge this. 
I will make a PR for docs next Monday.

PR #237, opened 2025-03-19
Simplify the creation of Docker/Singularity images

pzehner:
Given the recent security breach that affected the `tj-actions/changed-files` GitHub action, this was a good opportunity to reevaluate the way the project manages the creation of Docker images.

Currently, we have 2 sets of images: one for when the Dockerfiles are not modified (prefixed `_main`), and one for when the Dockerfiles are modified (prefixed `_pr`). Any feature branch that updated the Dockerfiles would then re-create its images on the fly. While very flexible, this process has not proven to be intensively used. Updating the Docker images is a somehow rare task. On the other hand, it generated a lot of complexity in the CI files.

This PR proposes to simplify the creation of Docker/Singularity images by separating the workflow between `create_images`, that only takes care of creating the images when a Dockerfile is modified, and the traditional `build_test`, that only uses the images. The registry will then only have one set of images, without suffixes.

If images need to be modified, then this should be done in a 2-step approach:

- Step 1: Create a PR that updates or creates the image; have it merged; and
- Step 2: Create a PR that uses the new image.

Progress so far:

- [x] Modify the CI files;
- [x] Force the creation of images without suffix;
- [x] Finalize the CI files.

### DIFF
--- a/.github/workflows/build_nightly.yaml
+++ b/.github/workflows/build_nightly.yaml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Configure Kokkos
         run: |
-          docker run -v ${{ github.workspace }}:/work ghcr.io/kokkos/kokkos-fft/base_${{ matrix.backend.image }}_main:latest \
+          docker run -v ${{ github.workspace }}:/work ghcr.io/kokkos/kokkos-fft/base_${{ matrix.backend.image }}:latest \
             cmake -B build_kokkos \
             -DCMAKE_INSTALL_PREFIX=/work/install \
             -DCMAKE_C_COMPILER=${{ matrix.backend.compiler.c }} \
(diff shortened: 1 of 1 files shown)


.github/workflows/create_base.yaml
```diff
@@ -79,19 +82,20 @@ jobs:
...
+            -t ghcr.io/kokkos/kokkos-fft/base_${{ matrix.image.name }}:latest \
+            --cache-from ghcr.io/kokkos/kokkos-fft/base_${{ matrix.image.name }}:latest \
             --build-arg BUILDKIT_INLINE_CACHE=1 \
             --progress=plain \
             docker/${{ matrix.image.name }}
 
       - name: Push Docker image
-        run: docker push ghcr.io/kokkos/kokkos-fft/base_${{ matrix.image.name }}_${{ inputs.image_suffix }}:${{ inputs.image_tag }}
+        run: docker push ghcr.io/kokkos/kokkos-fft/base_${{ matrix.image.name }}:latest
+        if: ${{ github.event != 'pull_request' }}
```
yasahi-hpc, line 93: Why do you prefer not to push Docker image in case of `pull_request`?

pzehner: Because doing so, you make all other PRs use the image of your current PR, which may be incomplete/faulty.

When a PR modifies the images, it pushes them when merged on `main`.

yasahi-hpc (approved):

PR #238, opened 2025-03-21
Introduce version check

yasahi-hpc:
This PR introduces a version consistency check among files and reference in `README.md`

- [x] Extract version information from `README.md` and some rst files
- [x] Extract version information from `CMakeLists.txt`
- [x] Raise an error in CI if some versions are different from those defined in `README.md`


tretre91 (approved):
Looks good to me

PR #241, opened 2025-03-28
Refactor unit-tests

yasahi-hpc:
This PR aims at refactoring unit-tests.

- [x] Unuse `Test_Types.hpp` which was used to define alias of `execution_space` and `View` types
- [x] Defining tests inside anonymous namespace
- [x] Make helper functions (`multiply`, `allclose`) to take `execution_space` instance argument. This modification is needed to run tests on non-default execution space in case device backend fft libs are disabled (see #239)
- [x] Rename test fixture name. Fixture names of compile time tests are now `CompileTest<Fixture_name>`


common/unit_test/Test_Extents.cpp
```diff
@@ -6,9 +6,10 @@
 #include <Kokkos_Random.hpp>
 #include <vector>
 #include "KokkosFFT_Extents.hpp"
-#include "Test_Types.hpp"
 #include "Test_Utils.hpp"
 
+namespace {
+using execution_space = Kokkos::DefaultExecutionSpace;
```
tpadioleau, line 12: Did you make sure it is allowed to define google tests inside anonymous namespaces ?

common/unit_test/Test_Utils.hpp
tpadioleau, line 1: Why do we use the raw pointers to access data ? At least we need to add assertions to make sure this is valid

common/unit_test/Test_Utils.hpp
```diff
@@ -39,15 +38,15 @@ bool allclose(const AViewType& a, const BViewType& b, double rtol = 1.e-5,
...
   const auto n = x.size();
   auto* ptr_x  = x.data();
 
   Kokkos::parallel_for(
       "KokkosFFT::Test::multiply",
-      Kokkos::RangePolicy<execution_space, Kokkos::IndexType<std::size_t>>{0,
-                                                                           n},
+      Kokkos::RangePolicy<ExecutionSpace, Kokkos::IndexType<std::size_t>>(exec,
+                                                                          0, n),
       KOKKOS_LAMBDA(const int& i) { ptr_x[i] = ptr_x[i] * a; });
```
tpadioleau, line 50:
```suggestion
      KOKKOS_LAMBDA(const std::size_t& i) { ptr_x[i] = ptr_x[i] * a; });
```

fft/unit_test/Test_Utils.hpp
tpadioleau, line 1: Some functions seem to be duplicated, would it be possible to share them between all tests ?

tpadioleau (changes requested):

common/unit_test/Test_Extents.cpp
yasahi-hpc: If I remembered, you encountered an issue with DDC. 
Do you remember what was that?

common/unit_test/Test_Utils.hpp
yasahi-hpc: It is certainly not a good practice.
At least, current APIs only accept Views with `LayoutLeft` or `LayoutRight`. Thus, `LayoutStrided` is not considered here. These helpers will be replaced with `testing-tools`

common/unit_test/Test_Utils.hpp
yasahi-hpc: Sure

fft/unit_test/Test_Utils.hpp
yasahi-hpc: That is true. I am planning to replace these with new `testing-tools` after Kokkos 5.0.

yasahi-hpc:
@tpadioleau 
Thank you for your review. I have updated accordingly.

common/unit_test/Test_Utils.hpp
tpadioleau: Sure.

common/unit_test/Test_Extents.cpp
tpadioleau: In DDC, we usually define a `struct` with the same name in different source files. We then define them in anonymous namespaces to avoid conflicts in the test executable.

fft/unit_test/Test_Utils.hpp
tpadioleau: I think we should keep track of this an issue.

common/unit_test/Test_Extents.cpp
yasahi-hpc: Do you mean defining a fixture with the same name in different source files?
Then, gtest will raise errors about the symbol collision, if we define tests inside anonymous namespace.

fft/unit_test/Test_Utils.hpp
yasahi-hpc: Sure

common/unit_test/Test_Extents.cpp
yasahi-hpc: Sure
Maybe it is safer to use normal namespaces?
I am using anonymous namespace in order to avoid the overlapping of fixture names in different files.

yasahi-hpc: Based on the discussion yesterday, I have moved `TEST_* marcos` outside an anonymous namespace.

PaulGannay (approved):
Looks good to me

yasahi-hpc:
@PaulGannay Thanks! I will merge this. Further improvements will be made with the new testing tools

PR #245, opened 2025-04-03
Update dependabot.yml

tpadioleau:


yasahi-hpc (approved):
Thank you for the fix

PR #249, opened 2025-04-04
Allow compilation while disabling device fft libraries

yasahi-hpc:
This PR aims at adding CMake options to enable/disable backend FFT libraries. If device backend is disabled, we execute the test and perf-test for FFT on `DefaultHostExecution` space.

- [x] Deprecate the `KokkosFFT_ENABLE_HOST_AND_DEVICE` option with warning
- [x] Library linking part is moved from `common/src/CMakeLists.txt`  to `fft/src/CMakeLists.txt` (this has been made in #251)
- [x] If device backend is disabled, only the example 5 (FFT on Host and Device example) is compiled.
- [x] Fix the random number generator in unit-tests and example 5 to run on `exec_space` instance (this has been made in #250)

Added 15/April
- [x] Do not allow to enable more than 1 device tpls


CMakeLists.txt
```diff
@@ -39,17 +39,23 @@ else()
 endif()
 
 # Device specific options
-set(HOST_BACKEND_ENABLED OFF CACHE BOOL "One of the Kokkos host backend is enabled")
-if(Kokkos_ENABLE_SERIAL OR Kokkos_ENABLE_THREADS OR Kokkos_ENABLE_OPENMP)
-  set(HOST_BACKEND_ENABLED ON)
+set(TO_ENABLE_FFTW ON CACHE BOOL "Whether to enable FFTW by default")
```
tpadioleau, line 42:
```suggestion
set(TO_ENABLE_FFTW ON)
```

Is there a specific reason to make it a cache variable ?

CMakeLists.txt
```diff
@@ -39,17 +39,23 @@ else()
...
-option(KokkosFFT_ENABLE_FFTW "Enable fftw as a KokkosFFT backend on CPUs" ${HOST_BACKEND_ENABLED})
+
+option(KokkosFFT_ENABLE_FFTW "Enable fftw as a KokkosFFT backend on CPUs" ${TO_ENABLE_FFTW})
 option(KokkosFFT_ENABLE_CUFFT "Enable cufft as a KokkosFFT backend on NVIDIA GPUs" ${Kokkos_ENABLE_CUDA})
 option(KokkosFFT_ENABLE_ROCFFT "Enable rocfft as a KokkosFFT backend on AMD GPUs" OFF)
 option(KokkosFFT_ENABLE_HIPFFT "Enable hipfft as a KokkosFFT backend on AMD GPUs" ${Kokkos_ENABLE_HIP})
 option(KokkosFFT_ENABLE_ONEMKL "Enable onemkl as a KokkosFFT backend on Intel GPUs" ${Kokkos_ENABLE_SYCL})
 
 # Set tpls
+set(KOKKOSFFT_HAS_DEFAULT_BACKEND TRUE)
```
tpadioleau, line 58: Why is `KOKKOSFFT_HAS_DEFAULT_BACKEND` not passed to `get_tpls_list` as the other variables ?

tpadioleau (changes requested):
What do you think of providing a Kokkos-fft execution space that aliases an execution space that is sure to have a corresponding fft backend ?

CMakeLists.txt
yasahi-hpc: Not really, I will modify accordingly

CMakeLists.txt
yasahi-hpc: I could not find an easy way to update a global variable through a function argument without directly manipulating it.
If there is a way, that is certainly better

yasahi-hpc:
Thank you for your review.

> What do you think of providing a Kokkos-fft execution space that aliases an execution space that is sure to have a corresponding fft backend ?

I would prefer to keep kokkos-fft as simple as possible. Thus, I would introduce a new execution space only if it is a necessary extension.

CMakeLists.txt
tpadioleau: Can you try https://cmake.org/cmake/help/latest/command/return.html ?

tpadioleau:
> I would prefer to keep kokkos-fft as simple as possible. Thus, I would introduce a new execution space only if it is a necessary extension.

Ok then can you consider defining a internal macro to avoid repeting this huge block:
```cpp
#if (defined(KOKKOSFFT_ENABLE_TPL_CUFFT) ||  \
     defined(KOKKOSFFT_ENABLE_TPL_ROCFFT) || \
     defined(KOKKOSFFT_ENABLE_TPL_HIPFFT) || \
     defined(KOKKOSFFT_ENABLE_TPL_ONEMKL))
```

CMakeLists.txt
yasahi-hpc: That is nicer.
But, it is supported only after CMake 3.25.0.
Can I just leave a comment to refactor this function as soon as we make the minimum CMake version 3.25.0.

yasahi-hpc:
> > I would prefer to keep kokkos-fft as simple as possible. Thus, I would introduce a new execution space only if it is a necessary extension.
> 
> Ok then can you consider defining a internal macro to avoid repeting this huge block:
> 
> ```c++
> #if (defined(KOKKOSFFT_ENABLE_TPL_CUFFT) ||  \
>      defined(KOKKOSFFT_ENABLE_TPL_ROCFFT) || \
>      defined(KOKKOSFFT_ENABLE_TPL_HIPFFT) || \
>      defined(KOKKOSFFT_ENABLE_TPL_ONEMKL))
> ```

Sure

yasahi-hpc:
@tpadioleau 
I have updated accordingly. May I have another review?

tpadioleau (commented):
Overall I am ok!

fft/unit_test/Test_Plans.cpp
```diff
@@ -84,24 +101,43 @@ void test_plan_constructible() {
   using PlanType =
       KokkosFFT::Plan<ExecutionSpace, RealView1DType, ComplexView1DType>;
 
-#if !defined(ENABLE_HOST_AND_DEVICE) &&                           \
-    (defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP) || \
-     defined(KOKKOS_ENABLE_SYCL))
-  // For GPUs without ENABLE_HOST_AND_DEVICE, a plan can be constructible from
-  // Kokkos::DefaultExecutionSpace only
+#if (defined(KOKKOSFFT_ENABLE_TPL_CUFFT) ||  \
```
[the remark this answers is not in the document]
yasahi-hpc: Sorry, I did not notice this. Fixed accordingly.

CMakeLists.txt
```diff
@@ -39,17 +39,23 @@ else()
 endif()
 
 # Device specific options
-set(HOST_BACKEND_ENABLED OFF CACHE BOOL "One of the Kokkos host backend is enabled")
-if(Kokkos_ENABLE_SERIAL OR Kokkos_ENABLE_THREADS OR Kokkos_ENABLE_OPENMP)
-  set(HOST_BACKEND_ENABLED ON)
+set(TO_ENABLE_FFTW ON)
```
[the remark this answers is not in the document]
yasahi-hpc: Sure. I have updated accordingly

yasahi-hpc:
@tpadioleau Thank you for the review again. I have updated. 
If you are good, I will merge this and release 0.3.0 soon

CMakeLists.txt
tpadioleau: I think if you apply the same strategy it should work `set(${default_backend_variable_name} ${has_default_backend} PARENT_SCOPE)`

yasahi-hpc: As you may imagine, that was my first implementation. 
Very surprisingly it did not update the variable in the parent scope.

tpadioleau: I see, let me just have a quick look before merging please

tpadioleau: If i do the following modification:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 616023c..8169805 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,11 +55,13 @@ option(KokkosFFT_ENABLE_HIPFFT "Enable hipfft as a KokkosFFT backend on AMD GPUs
 option(KokkosFFT_ENABLE_ONEMKL "Enable onemkl as a KokkosFFT backend on Intel GPUs" ${Kokkos_ENABLE_SYCL})
 
 # Set tpls
-set(KOKKOSFFT_HAS_DEFAULT_BACKEND TRUE)
+set(KOKKOSFFT_HAS_DEFAULT_BACKEND)
 set(KOKKOSFFT_TPL_LIST)
 set(KOKKOSFFT_BACKEND_LIST)
 include(KokkosFFT_tpls)
-get_tpls_list(KOKKOSFFT_TPL_LIST KOKKOSFFT_BACKEND_LIST)
+message(STATUS "Before get_tpls_list \"${KOKKOSFFT_HAS_DEFAULT_BACKEND}\"")
+get_tpls_list(KOKKOSFFT_TPL_LIST KOKKOSFFT_BACKEND_LIST KOKKOSFFT_HAS_DEFAULT_BACKEND)
+message(STATUS "After get_tpls_list \"${KOKKOSFFT_HAS_DEFAULT_BACKEND}\"")
 
 # ==================================================================
 # CMake Summary
diff --git a/cmake/KokkosFFT_tpls.cmake b/cmake/KokkosFFT_tpls.cmake
index 0385022..cd05750 100644
--- a/cmake/KokkosFFT_tpls.cmake
+++ b/cmake/KokkosFFT_tpls.cmake
@@ -11,7 +11,7 @@
 # \param tpls_list[inout] A list of tpls that are needed
 # \param backend_list[inout] A list of backends corresponding to tpls
 #        backend_list is used to display
-function(get_tpls_list tpls_list backend_list)
+function(get_tpls_list tpls_list backend_list default_backend)
   set(KOKKOSFFT_HAS_DEVICE FALSE)
   set(has_default_backend TRUE)
   if(Kokkos_ENABLE_CUDA)
@@ -83,5 +83,5 @@ function(get_tpls_list tpls_list backend_list)
   set(${backend_list} ${${backend_list}} PARENT_SCOPE)
   # FIXME
   # We will use "return" to update this value as soon as we make the minimum CMake version 3.25.0
-  set(KOKKOSFFT_HAS_DEFAULT_BACKEND ${has_default_backend} PARENT_SCOPE)
+  set(${default_backend} ${has_default_backend} PARENT_SCOPE)
 endfunction()
```

I get the following output
```
[main] Configuring project: kokkos-fft 
[proc] Executing command: /local/home/tpadiole/spack-0.23.1/var/spack/environments/kokkos-fft-cpu/.spack-env/view/bin/cmake -DCMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_C_COMPILER:FILEPATH=/usr/bin/clang-20 -DCMAKE_CXX_COMPILER:FILEPATH=/usr/bin/clang++-20 --no-warn-unused-cli -S/local/home/tpadiole/Codes/kokkos-fft -B/local/home/tpadiole/Codes/kokkos-fft/build -G Ninja
[cmake] Not searching for unused variables given on the command line.
[cmake] -- Enabled Kokkos devices: SERIAL
[cmake] -- Before get_tpls_list ""
[cmake] -- After get_tpls_list "TRUE"
[cmake] -- KokkosFFT version: 0.2.1
[cmake] -- KokkosFFT TPLs:
[cmake] --       Host Serial: FFTW_SERIAL
[cmake] 
[cmake] 
[cmake] -- Configuring done (0.1s)
[cmake] -- Generating done (0.0s)
[cmake] -- Build files have been written to: /local/home/tpadiole/Codes/kokkos-fft/build
```

tpadioleau (changes requested):
I have some final comments/questions:
- I see we still have some `ENABLE_HOST_AND_DEVICE` in the sources, is it expected ? Do we still need them ?
- We need to update the documentation that mentions `KokkosFFT_ENABLE_HOST_AND_DEVICE`. We can document it as deprecated.

CMakeLists.txt
yasahi-hpc: Thanks! This works

I have tried the following which does not work
`set(${has_default_backend} ${${has_default_backend}} PARENT_SCOPE)`

yasahi-hpc:
> I have some final comments/questions:
> 
> * I see we still have some `ENABLE_HOST_AND_DEVICE` in the sources, is it expected ? Do we still need them ?
> * We need to update the documentation that mentions `KokkosFFT_ENABLE_HOST_AND_DEVICE`. We can document it as deprecated.

I am planning to do the followings in the next PR
1. Remove `ENABLE_HOST_AND_DEVICE` in the sources
2. Update documentation
3. Unuse `KokkosFFT_ENABLE_HOST_AND_DEVICE` in CI

tpadioleau (approved):
Looks to me!

yasahi-hpc:
Thanks! I will merge this

PR #250, opened 2025-04-07
Run fill_random on a given execution space instance

yasahi-hpc:
This PR aims at running fill_random on a given execution space instance which is needed for #249

### DIFF
--- a/fft/unit_test/Test_Transform.cpp
+++ b/fft/unit_test/Test_Transform.cpp
@@ -46,7 +46,7 @@ void test_fft1_identity(T atol = 1.0e-12) {
   using RealView1DType = Kokkos::View<T*, LayoutType, execution_space>;
   using ComplexView1DType =
       Kokkos::View<Kokkos::complex<T>*, LayoutType, execution_space>;
-
+  execution_space exec;
   for (int i = 1; i < maxlen; i++) {
     ComplexView1DType a("a", i), a_ref("a_ref", i);
     ComplexView1DType a_hat("a_hat", i), inv_a_hat("inv_a_hat", i);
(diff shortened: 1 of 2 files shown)


PaulGannay (approved):
Looks good to me

fft/unit_test/Test_Transform.cpp
```diff
@@ -295,28 +299,26 @@ void test_fft1_1dfft_1dview() {
...
+  KokkosFFT::fft(exec, x, out);  // default: KokkosFFT::Normalization::backward
+  KokkosFFT::fft(exec, x, out_b, KokkosFFT::Normalization::backward);
+  KokkosFFT::fft(exec, x, out_o, KokkosFFT::Normalization::ortho);
+  KokkosFFT::fft(exec, x, out_f, KokkosFFT::Normalization::forward);
 
-  fft1(execution_space(), x, ref);
-  multiply(execution_space(), out_o, sqrt(static_cast<T>(len)));
-  multiply(execution_space(), out_f, static_cast<T>(len));
+  fft1(exec, x, ref);
+  multiply(exec, out_o, sqrt(static_cast<T>(len)));
```
tpadioleau, line 314: Not related to this PR but I see an unqualified call to `sqrt`, aren't we calling the C version in this case ?

tpadioleau (dismissed):
Ok for me, to make it even more uniform we could also pass `exec` to `Kokkos::deep_copy`.

yasahi-hpc:
@PaulGannay 
Thanks! I will merge this

fft/unit_test/Test_Transform.cpp
yasahi-hpc: Oh, that is true. 
We should use `std::sqrt` or `Kokkos::sqrt`

tpadioleau: We can `grep` later and change this in a new PR.

yasahi-hpc: I made it in this PR. After CI passes, I will merge this

tpadioleau: ok

PR #251, opened 2025-04-07
Linking library logic is moved from common/src/CMake to fft/src/CMake

yasahi-hpc:
This PR aims at moving the library linking part from `common/src/CMake` to `fft/src/CMake` which is needed in PR #249


yasahi-hpc:
Firstly, the functions under common are completely independent from FFT libraries. Thus, we should not link libraries under common to make common functions standalone. 

Second, the logic of `is_AllowedSpace` is more related to the available backend libraries. This logic changes based on the enabled libraries which should not visible from common traits. (See coming changes in #249)

tpadioleau:
> Firstly, the functions under common are completely independent from FFT libraries. Thus, we should not link libraries under common to make common functions standalone.
> 
> Second, the logic of `is_AllowedSpace` is more related to the available backend libraries. This logic changes based on the enabled libraries which should not visible from common traits. (See coming changes in #249)

So if I undersand correctly, we don't want `common` to depend on the backends enabled ?

tpadioleau (approved):
Seems good to me

yasahi-hpc:
> 

That is true. We want `common` functions to run on any `Kokkos` backends.

yasahi-hpc:
@tpadioleau 

Thank you for your approval. I will merge this

PR #258, opened 2025-04-15
Fixes a link in the documentation

tpadioleau:
The link https://kokkos.org/kokkos-core-wiki/requirements.html seems now broken, did we mean https://kokkos.org/kokkos-core-wiki/get-started/requirements.html ?

### DIFF
--- a/docs/intro/building.rst
+++ b/docs/intro/building.rst
@@ -15,7 +15,7 @@ Available CMake options for kokkos-fft are listed.
 Compiler versions
 -----------------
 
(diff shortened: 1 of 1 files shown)


yasahi-hpc (approved):
Thank you for the fix. 
I did not notice that. 

Do you think it make sense to introduce a CI to check links?

tpadioleau:
> Do you think it make sense to introduce a CI to check links?

Yes it could help detecting broken links

PR #262, opened 2025-04-21
Introduce a helper to convert a view type with different base value type

yasahi-hpc:
This PR aims at introducing a helper to define a new view type with different value type while keeping other properties.
This helper is useful to allocate a buffer in complex data from the same shape of real data.

```C++
using DerivedViewType2 = typename KokkosFFT::Impl::ConvertedViewType<ViewType1, T2>::type;
```

- [x] Adding a `add_pointer_n` helper to manipulate pointer types
- [x] Adding a `ConvertedViewType` helper for base value type conversion


PaulGannay (approved):
Looks good to me

yasahi-hpc:
@PaulGannay Thanks! I will merge this

PR #264, opened 2025-04-25
Extend View to View matcher up to 8D

yasahi-hpc:
This PR aims at extending View to View matcher up to 8D

- [x] 3D to 8D versions of `ViewErrors` and `FindErrors` are implemented
- [x] Unit-tests are added on 3D-8D Views

It may be counter-intuitive to give the global index from `LayoutLeft` order. 
This means that value at `(1, 1, 2, 2)` is printed before value at `(0, 1, 1, 3)`

```C++
  // [NOTE] Order reversed because of sorting by global idx
  // "Mismatched elements (by indices):\n"
  // "  Index (1, 1, 2, 2): actual 4.0000000000000000 vs expected 3.0000000000000000 (diff=1.0000000000000000)\n"
  // "  Index (0, 1, 1, 3): actual 3.0000599999999999 vs expected 3.0000000000000000 (diff=0.0000599999999999)";
  // NOLINTEND(*)
```

### DIFF
--- a/testing/src/KokkosFFT_CountErrors.hpp
+++ b/testing/src/KokkosFFT_CountErrors.hpp
@@ -57,8 +57,7 @@ Kokkos::Iterate get_iteration_order(const ViewType& view) {
 /// \tparam Layout The layout type of the Kokkos views.
 /// \tparam iType The integer type used for indexing the view elements.
 template <KokkosExecutionSpace ExecutionSpace, KokkosView AViewType,
-          KokkosView BViewType, KokkosLayout Layout, std::size_t Rank,
-          typename iType>
+          KokkosView BViewType, KokkosLayout Layout, typename iType>
 struct ViewErrors {
  private:
   // Since MDRangePolicy is not available for 7D and 8D views, we need to
@@ -205,8 +204,7 @@ struct ViewErrors {
 /// \tparam Layout The memory layout type of the Kokkos views.
 /// \tparam iType The integer type used for indexing the view elements.
 template <KokkosExecutionSpace ExecutionSpace, KokkosView AViewType,
-          KokkosView BViewType, KokkosLayout Layout, std::size_t Rank,
-          typename iType>
+          KokkosView BViewType, KokkosLayout Layout, typename iType>
 struct FindErrors {
  private:
   // Since MDRangePolicy is not available for 7D and 8D views, we need to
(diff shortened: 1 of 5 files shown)


testing/src/KokkosFFT_CountErrors.hpp
```diff
@@ -79,118 +80,1308 @@ struct ViewErrors<ExecutionSpace, AViewType, BViewType, Layout, 1, iType> {
...
+  /// views.
+  /// \param err [in,out] The error counter incremented if a mismatch is
+  /// detected.
+  KOKKOS_INLINE_FUNCTION
+  void operator()(const iType& i0, const iType& i1, std::size_t& err) const {
+    auto tmp_a     = m_a(i0, i1);
+    auto tmp_b     = m_b(i0, i1);
+    bool not_close = are_not_close(tmp_a, tmp_b, m_rtol, m_atol);
+    err += static_cast<std::size_t>(not_close);
+  };
```
PaulGannay, line 186: This could also be factorised into `ErrorViewBase`, it becomes complicated due to the 7 and 8 case though.

testing/src/KokkosFFT_CountErrors.hpp
```diff
@@ -79,118 +80,1308 @@ struct ViewErrors<ExecutionSpace, AViewType, BViewType, Layout, 1, iType> {
...
+  AErrorViewType m_a_error;
+  BErrorViewType m_b_error;
+  CountViewType m_loc_error;
+  CountType m_count;
+
+  double m_rtol;
+  double m_atol;
+
+  using policy_type =
+      Kokkos::RangePolicy<ExecutionSpace, Kokkos::IndexType<iType>>;
```
PaulGannay, line 738: Same comment as for ViewErrors, this can be factorised into a base class.

PaulGannay (changes requested):
ViewErrors and FindErrors could be factorised to reduce code reuse between the different dimensions

yasahi-hpc:
@PaulGannay Thank you for your review. I will update accordingly.

testing/src/KokkosFFT_CountErrors.hpp
science-enthusiast: If `operator()` is going to be implemented as a variadic function template, the number of arguments apart from `err` being 7 and 8 can be dealt with by imposing constraints on the template either with SFINAE or with a requires clause.

template <typename... Args, typename = std::enable_if_t<(sizeof...(Args) != 7) && (sizeof...(Args) != 8)>>
void operator(std::size_t& err, Args... args) { ... }

or

template <typename... Args>
requires (sizeof...(Args) != 7) && (sizeof...(Args) != 8)
void operator(std::size_t& err, Args... args) { ... }

yasahi-hpc:
@PaulGannay I have updated accordingly. Can I have another review

PaulGannay (approved):
Very good changes, I have two remarks though (I won't block the merge over any of this):
 1 - I wonder if you should follow @science-enthusiast advices and use `std::enable_if` or `require` instead of doing all the `if constexpr` ?
For the smaller functions (`get_policy` and `CountErrors::operator()`) it stays readable as is, but for `FindErrorsInternal::operator()` and `get_global_idx` it may be better to have the logic split into different smaller functions? If you do agree to change it for the big function, I think it would be better to make the change for all function to stay consistent?

 2 - Also, I wonder what happens if we get a view with rank >= 9? I know it is not possible as of now, but if it ever becomes possible, will it fail gracefully?

yasahi-hpc:
> Very good changes, I have two remarks though (I won't block the merge over any of this): 1 - I wonder if you should follow @science-enthusiast advices and use `std::enable_if` or `require` instead of doing all the `if constexpr` ? For the smaller functions (`get_policy` and `CountErrors::operator()`) it stays readable as is, but for `FindErrorsInternal::operator()` and `get_global_idx` it may be better to have the logic split into different smaller functions? If you do agree to change it for the big function, I think it would be better to make the change for all function to stay consistent?

Actually, I am not sure how to split the operator into different smaller functions (`get_global_idx` can be split). 
If we template these operators, it fails to be deduced by MDRange.
If we can use recursion, `get_global_idx` can be further simplified, but it is almost always a bad idea to use recursion on device. If you think is better, I can split `get_global_idx` function into smaller ones.
 
> 2 - Also, I wonder what happens if we get a view with rank >= 9? I know it is not possible as of now, but if it ever becomes possible, will it fail gracefully?

I think we will just update the code then.

PaulGannay:
> If we can use recursion, `get_global_idx` can be further simplified, but it is almost always a bad idea to use recursion on device. If you think is better, I can split `get_global_idx` function into smaller ones.

Recursion would be done over the rank, so is doable at compile time, and shouldn't affect runtime performances? This way, you'd have no need to split anything?

yasahi-hpc:
> > If we can use recursion, `get_global_idx` can be further simplified, but it is almost always a bad idea to use recursion on device. If you think is better, I can split `get_global_idx` function into smaller ones.
> 
> Recursion would be done over the rank, so is doable at compile time, and shouldn't affect runtime performances? This way, you'd have no need to split anything?

I can try. May be OK for the compile time recursion. 
Recursion on device frequently results in runtime errors not even lower performance.

yasahi-hpc:
@PaulGannay I think we can do that even without recursion which is safer on device

PaulGannay (approved):
Nice change, indeed, no need of recursion.

yasahi-hpc:
Current CI failure is unseated. I will merge this. Thank you for your review @PaulGannay

PR #265, opened 2025-04-25
Add ulps comparison operator

yasahi-hpc:
This PR introduces a comparison operator for unit-testing.
Some companions of `half_t` and `bhalf_t` are suspicious. 
In addition, `T c = -Kokkos::Experimental::infinity<T>::value;` does not compile for `half_t` and `bhalf_t`.
Should we say that positive zero is not equal to negative zero?

- [x] Implement a function that checks whether two floating point values are close in terms of ulps
- [x] Unit tests for positive, negative, near-zeros, infinities and nans.

Edited on 28/April
- [x] Adding `nextafter` helper for `fp16`
- [x]  Issues on AMD GPUs. Looks like `fp16` is `float` there.

```
[ 77%] Building CXX object testing/unit_test/CMakeFiles/unit-tests-kokkos-fft-testing.dir/Test_AlmostEqualUlps.cpp.o
/work/testing/unit_test/Test_AlmostEqualUlps.cpp:44:29: error: no matching function for call to 'bit_cast'
  std::uint16_t uint_from = Kokkos::bit_cast<std::uint16_t>(from);
                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/work/testing/unit_test/Test_AlmostEqualUlps.cpp:102:12: note: in instantiation of function template specialization '(anonymous namespace)::nextafter_fp16<float>' requested here
    return nextafter_fp16<T>(from, to);
           ^
/work/testing/unit_test/Test_AlmostEqualUlps.cpp:303:3: note: in instantiation of function template specialization '(anonymous namespace)::test_almost_equal_ulps_positive<float, Kokkos::View<bool, Kokkos::HIP>>' requested here
  test_almost_equal_ulps_positive<float_type, BoolViewType>();
```

Edited on 30/April
It seems that `half_t` is a half type but `bhalf_t` is not. The following code can work on HIP backend.

```C++
static_assert(sizeof(Kokkos::Experimental::half_t) == 2);
static_assert(sizeof(Kokkos::Experimental::bhalf_t) == 4);
// On CUDA backend, the following compiles
//static_assert(sizeof(Kokkos::Experimental::bhalf_t) == 2);
```

### DIFF
--- a/testing/src/KokkosFFT_Ulps.hpp
+++ b/testing/src/KokkosFFT_Ulps.hpp
@@ -1,0 +1,79 @@
+// SPDX-FileCopyrightText: (C) The kokkos-fft development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+#ifndef KOKKOSFFT_ULPS_HPP
+#define KOKKOSFFT_ULPS_HPP
+
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Half.hpp>
+#include <Kokkos_MathematicalFunctions.hpp>
+#include <Kokkos_BitManipulation.hpp>
+#include "KokkosFFT_Concepts.hpp"
+
+namespace KokkosFFT {
+namespace Testing {
+namespace Impl {
+
+// Helper structure to select the correct integer type for bit reinterpretation
+template <typename T>
+struct FloatIntMap;
+
+template <>
+struct FloatIntMap<float> {
+  using IntType = std::int32_t;
+};
+template <>
+struct FloatIntMap<double> {
+  using IntType = std::int64_t;
+};
+
+template <>
+struct FloatIntMap<Kokkos::Experimental::half_t> {
+  using IntType = std::int16_t;
+};
+template <>
+struct FloatIntMap<Kokkos::Experimental::bhalf_t> {
+  using IntType = std::int16_t;
+};
+
+template <typename ScalarA, typename ScalarB>
+KOKKOS_INLINE_FUNCTION bool almost_equal_ulps(ScalarA a, ScalarB b,
+                                              std::size_t max_ulps_diff) {
+  // Handle non-finite cases and exact equality first
+  if (Kokkos::isnan(a) || Kokkos::isnan(b)) return false;
+
+  // Exact equality check: Handles +0 == -0, inf == inf, -inf == -inf
+  if (a == b) return true;
+
+  // Since a != b, we only need to check if a or b is infinite
+  if (Kokkos::isinf(a) || Kokkos::isinf(b)) return false;
+
+  // Get the integer representation of the floats
+  // Not sure how to compare if two types do not have the common type
+  using CommonFloatType = std::common_type_t<ScalarA, ScalarB>;
+  using IntType         = typename FloatIntMap<CommonFloatType>::IntType;
+
+  // Reinterpret the bits using Kokkos::bit_cast
+  IntType int_a = Kokkos::bit_cast<IntType>(a);
+  IntType int_b = Kokkos::bit_cast<IntType>(b);
+
+  // Calculate the ULP difference
+  IntType ulps_diff = int_a - int_b;
+
+  // Do not know how Kokkos::abs works on int16_t,
+  // so hardcode the absolute value
+  using UnsignedIntType         = typename std::make_unsigned<IntType>::type;
+  UnsignedIntType abs_ulps_diff = (ulps_diff < 0)
+                                      ? static_cast<UnsignedIntType>(-ulps_diff)
+                                      : static_cast<UnsignedIntType>(ulps_diff);
+
+  // Compare with the maximum allowed ULP difference
+  return abs_ulps_diff <= static_cast<UnsignedIntType>(max_ulps_diff);
+}
+
+}  // namespace Impl
+}  // namespace Testing
+}  // namespace KokkosFFT
+
+#endif
--- a/testing/src/KokkosFFT_Ulps.hpp
+++ b/testing/src/KokkosFFT_Ulps.hpp
@@ -63,7 +63,7 @@ KOKKOS_INLINE_FUNCTION UIntType float_to_biased_int(FloatType from) {
   // e.g., 0x80000000 for int32_t, 0x8000 for int16_t
   constexpr UIntType sign_mask = UIntType(1) << (sizeof(UIntType) * 8 - 1);
 
-  if ((to & sign_mask) != 0) {
+  if (to & sign_mask) {
     // Original float was negative (or -0.0f)
     // Bitwise NOT to reverse order and map to lower half conceptually
     return ~to + 1;
--- a/testing/src/KokkosFFT_Ulps.hpp
+++ b/testing/src/KokkosFFT_Ulps.hpp
@@ -21,34 +21,50 @@ struct FloatIntMap;
 
 template <>
 struct FloatIntMap<float> {
-  using IntType = std::int32_t;
+  using IntType = std::uint32_t;
 };
 template <>
 struct FloatIntMap<double> {
-  using IntType = std::int64_t;
+  using IntType = std::uint64_t;
 };
 
 // half_t and bhalf_t can be alias to float types
 #if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
 template <>
 struct FloatIntMap<Kokkos::Experimental::half_t> {
-  using IntType = std::int16_t;
+  using IntType = std::uint16_t;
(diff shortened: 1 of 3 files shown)


testing/unit_test/Test_AlmostEqualUlps.cpp
```diff
@@ -0,0 +1,327 @@
...
+  using float_type   = T;
+  using BoolViewType = Kokkos::View<bool, execution_space>;
+};
+
+// Helper function for nextafter on fp16 types
+template <typename fp16_t>
+  requires(std::is_same_v<fp16_t, Kokkos::Experimental::half_t> ||
+           std::is_same_v<fp16_t, Kokkos::Experimental::bhalf_t>)
+fp16_t nextafter_fp16(fp16_t from, fp16_t to) {
+  constexpr std::uint16_t FP16_SIGN_MASK = 0x8000;
```
PaulGannay, line 27: In kokkos, there is an `exponent_mask` and a `fraction_mask` defined:
[```
#ifdef KOKKOS_IMPL_HALF_TYPE_DEFINED
template <>
inline constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
    exponent_mask<Kokkos::Experimental::half_t>{0b0'11111'0000000000};
template <>
inline constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
    fraction_mask<Kokkos::Experimental::half_t>{0b0'00000'1111111111};
#endif

#ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
template <>
inline constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
    exponent_mask<Kokkos::Experimental::bhalf_t>{0b0'11111111'0000000};
template <>
inline constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
    fraction_mask<Kokkos::Experimental::bhalf_t>{0b0'00000000'1111111};
#endif
```](https://github.com/kokkos/kokkos/blob/be496a754c6518e2cbec1027f84f45f423336638/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp#L258)

File: testing/src/KokkosFFT_Ulps.hpp [hunk 2]
```diff
@@ -0,0 +1,85 @@
...
+struct FloatIntMap<Kokkos::Experimental::bhalf_t> {
+  using IntType = std::int16_t;
+};
+#endif
+
+template <typename ScalarA, typename ScalarB>
+KOKKOS_INLINE_FUNCTION bool almost_equal_ulps(ScalarA a, ScalarB b,
+                                              std::size_t max_ulps_diff) {
+  // Handle non-finite cases and exact equality first
+  if (Kokkos::isnan(a) || Kokkos::isnan(b)) return false;
```
PaulGannay, line 50: If both are nan, it should return true, right?

testing/src/KokkosFFT_Ulps.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+  // Exact equality check: Handles +0 == -0, inf == inf, -inf == -inf
+  if (a == b) return true;
+
+  // Since a != b, we only need to check if a or b is infinite
+  if (Kokkos::isinf(a) || Kokkos::isinf(b)) return false;
+
+  // Get the integer representation of the floats
+  // Not sure how to compare if two types do not have the common type
```
PaulGannay, line 59: This is were [the method used in Kokkos](https://github.com/kokkos/kokkos/blob/be496a754c6518e2cbec1027f84f45f423336638/core/unit_test/TestMathematicalFunctions.hpp#L302) might be good: it will work despite the type difference.

testing/src/KokkosFFT_Ulps.hpp
yasahi-hpc: I can change the behavior. Current implementation comes from the article 
https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

It says that

> NaNs – two NaNs could have very similar or even identical representations, but they are not supposed to compare as equal

testing/src/KokkosFFT_Ulps.hpp
yasahi-hpc: I am not quite sure how to compare ulps in different precisions. Is there a common way to do that?

testing/src/KokkosFFT_Ulps.hpp
PaulGannay: In the general case, sure, but this is a test function, right?
So if the reference function returns `NaN`, the function you are testing should also return `NaN` and your test function should return `true`.

yasahi-hpc: OK

testing/unit_test/Test_AlmostEqualUlps.cpp
```diff
@@ -0,0 +1,342 @@
...
+    return Kokkos::bit_cast<fp16_t>((to > from) ? FP16_SMALLEST_POS_DN
+                                                : FP16_SMALLEST_NEG_DN);
+  }
+  if (uint_from == FP16_NEG_ZERO) {  // from is -0.0
+    // Return smallest magnitude number with the sign of 'to'.
+    // Standard nextafter(-0, negative) -> smallest_negative
+    // And nextafter(-0, positive) -> smallest_positive
+    return Kokkos::bit_cast<fp16_t>((to > from) ? FP16_SMALLEST_POS_DN
+                                                : FP16_SMALLEST_NEG_DN);
+  }
```
PaulGannay, line 62: Same comment as the other PR, can be merged with `from == fp16_t(0)`.

testing/src/KokkosFFT_Ulps.hpp
yasahi-hpc: According to [this](https://stackoverflow.com/questions/1565164/what-is-the-rationale-for-all-comparisons-returning-false-for-ieee754-nan-values), `NAN == NAN` should be `false`. So, I reverted to the original implementation. 

This is also confirmed in the gtest implementation. See https://github.com/google/googletest/blob/e9092b12dc3cf617d47578f13a1f64285cfa5b2f/googletest/include/gtest/internal/gtest-internal.h#L336-L342

PaulGannay: `NaN==NaN` should be `false` in the general case (when emulating operator==() for instance), but it is counter productive when writing a test function: if someone wants to test a View containing expected NaN values, he won't be able to use this function to do so.

If you prefer to align with gtest, I won't block the merge over this but the function would be much more useful with `NaN==NaN` returning `true`.

The gtest function is usable because it only works on scalar value, were it to work on an array it would face the same problem.

yasahi-hpc: I prefer to align with gtest at least as a default.
For Nan to Nan comparison purpose, I would rather like to implement different function or add an option where Nan==Nan is considered as `true`.

testing/src/KokkosFFT_Ulps.hpp
```diff
@@ -0,0 +1,97 @@
...
+  using CommonFloatType = std::common_type_t<ScalarA, ScalarB>;
+  using UIntType        = typename FloatIntMap<CommonFloatType>::IntType;
+
+  // Reinterpret the bits using Kokkos::bit_cast
```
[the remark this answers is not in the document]
yasahi-hpc: Sure, I forgot to remove the comments

testing/src/KokkosFFT_Ulps.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+template <typename UIntType, typename FloatType>
+  requires(sizeof(UIntType) == sizeof(FloatType))
+KOKKOS_INLINE_FUNCTION UIntType float_to_biased_int(FloatType from) {
+  UIntType to = Kokkos::bit_cast<UIntType>(from);
+
+  // e.g., 0x80000000 for int32_t, 0x8000 for int16_t
+  constexpr UIntType sign_mask = UIntType(1) << (sizeof(UIntType) * 8 - 1);
+
+  if ((to & sign_mask) != 0) {
```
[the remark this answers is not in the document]
yasahi-hpc: In my opinion, it is better to rely on bit operations for accuracy. This way, we do not need to rely on the implementation details of fp16 in Kokkos.
I am not 100 % sure whether a comparison operator for float types works in the expected way.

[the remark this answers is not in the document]
yasahi-hpc: It may be that 
`if (to & sign_mask)`
is sufficient

yasahi-hpc:
@PaulGannay 
I think I have addressed all the points except for NaN equality political decision.

May I have another review please

yasahi-hpc:
@PaulGannay Thank you for approval. I will merge this

PR #269, opened 2025-04-29
Cartesian product of types

yasahi-hpc:
This PR aims at introducing an API to generate cartesian product of types which originally developed for [DDC](https://github.com/CExA-project/ddc). For example, the following types are identical. This helper is used for [TYPED_TEST](https://github.com/google/googletest/blob/main/docs/advanced.md#typed-tests).

```C++
using cartesian_product_type =
       KokkosFFT::Testing::make_cartesian_types<std::tuple<float, double>, std::tuple<float, double>>;

using reference_type = testing::Types<std::tuple<float, float>, std::tuple<double, float>,
                      std::tuple<float, double>, std::tuple<double, double>;
```

- [x] Introduce an API `make_cartesian_types` to generate cartesian product of types
- [x] Compile time unit-tests for type generators

### DIFF
--- a/testing/src/KokkosFFT_TypeCartesianProduct.hpp
+++ b/testing/src/KokkosFFT_TypeCartesianProduct.hpp
@@ -1,0 +1,104 @@
+// SPDX-FileCopyrightText: (C) The kokkos-fft development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+#ifndef KOKKOSFFT_TYPE_CARTESIAN_PRODUCT_HPP
+#define KOKKOSFFT_TYPE_CARTESIAN_PRODUCT_HPP
+
+#include <tuple>
+#include <type_traits>
+
+namespace KokkosFFT {
+namespace Testing {
+namespace Impl {
+
+/// Transform a sequence S to a tuple:
+/// - a std::integer_sequence<T, Ints...> to a
+/// std::tuple<std::integral_constant<T, Ints>...>
+/// - a std::pair<T, U> to a std::tuple<T, U>
+/// - identity otherwise (std::tuple)
+template <class S>
+struct to_tuple {
+  using type = S;
+};
+
+template <class T, T... Ints>
+struct to_tuple<std::integer_sequence<T, Ints...>> {
+  using type = std::tuple<std::integral_constant<T, Ints>...>;
+};
+
+template <class T, class U>
+struct to_tuple<std::pair<T, U>> {
+  using type = std::tuple<T, U>;
+};
+
+template <class S>
+using to_tuple_t = typename to_tuple<S>::type;
+
+template <class TupleOfTuples, class Tuple>
+struct for_each_tuple_cat;
+
+template <class... Tuples, class Tuple>
+struct for_each_tuple_cat<std::tuple<Tuples...>, Tuple> {
+  using type = std::tuple<decltype(std::tuple_cat(
+      std::declval<Tuples>(), std::declval<std::tuple<Tuple>>()))...>;
+};
+
+template <class... Tuples, class Tuple>
+struct for_each_tuple_cat<std::tuple<Tuples...>, std::tuple<Tuple>> {
+  using type = std::tuple<decltype(std::tuple_cat(
+      std::declval<Tuples>(), std::declval<std::tuple<Tuple>>()))...>;
+};
+
+/// Construct a tuple of tuples that is the result of the concatenation of the
+/// tuples in TupleOfTuples with Tuple.
+template <class TupleOfTuples, class Tuple>
+using for_each_tuple_cat_t =
+    typename for_each_tuple_cat<TupleOfTuples, Tuple>::type;
+
+template <class InTupleOfTuples, class OutTupleOfTuples>
+struct cartesian_product_impl;
+
+template <class... HeadArgs, class... TailTuples, class OutTupleOfTuples>
+struct cartesian_product_impl<
+    std::tuple<std::tuple<HeadArgs...>, TailTuples...>, OutTupleOfTuples>
+    : cartesian_product_impl<
+          std::tuple<TailTuples...>,
+          decltype(std::tuple_cat(
+              std::declval<for_each_tuple_cat_t<OutTupleOfTuples,
+                                                std::tuple<HeadArgs>>>()...))> {
+};
+
+template <class OutTupleOfTuples>
+struct cartesian_product_impl<std::tuple<>, OutTupleOfTuples> {
+  using type = OutTupleOfTuples;
+};
+
+/// Generate a std::tuple cartesian product from multiple tuple-like structures
+/// (std::tuple, std::integer_sequence and std::pair) Do not rely on the
+/// ordering result.
+template <class... InTuplesLike>
+using cartesian_product_t =
+    typename cartesian_product_impl<std::tuple<to_tuple_t<InTuplesLike>...>,
+                                    std::tuple<std::tuple<>>>::type;
+
+/// Transform a std::tuple<Args...> to a testing::Types<Args...>, identity
+/// otherwise
+template <class T>
+struct tuple_to_types {
+  using type = T;
+};
+
+template <class... Args>
+struct tuple_to_types<std::tuple<Args...>> {
+  using type = testing::Types<Args...>;
+};
+
+template <class T>
+using tuple_to_types_t = typename tuple_to_types<T>::type;
+
+}  // namespace Impl
+}  // namespace Testing
+}  // namespace KokkosFFT
+
+#endif
--- a/testing/src/KokkosFFT_TypeCartesianProduct.hpp
+++ b/testing/src/KokkosFFT_TypeCartesianProduct.hpp
@@ -98,6 +98,12 @@ template <class T>
 using tuple_to_types_t = typename tuple_to_types<T>::type;
 
 }  // namespace Impl
+
+// API to generate caresian product of input types
+template <class... InTuplesLike>
+using make_cartesian_types =
+    Impl::tuple_to_types_t<Impl::cartesian_product_t<InTuplesLike...>>;
+
 }  // namespace Testing
 }  // namespace KokkosFFT
 
--- a/testing/src/KokkosFFT_TypeCartesianProduct.hpp
+++ b/testing/src/KokkosFFT_TypeCartesianProduct.hpp
@@ -35,30 +35,38 @@ struct to_tuple<std::pair<T, U>> {
 template <class S>
 using to_tuple_t = typename to_tuple<S>::type;
 
-template <class TupleOfTuples, class Tuple>
+template <class TupleOfTuples, class Type>
 struct for_each_tuple_cat;
 
-template <class... Tuples, class Tuple>
-struct for_each_tuple_cat<std::tuple<Tuples...>, Tuple> {
+template <class... Tuples, class Type>
+struct for_each_tuple_cat<std::tuple<Tuples...>, Type> {
   using type = std::tuple<decltype(std::tuple_cat(
-      std::declval<Tuples>(), std::declval<std::tuple<Tuple>>()))...>;
+      std::declval<Tuples>(), std::declval<std::tuple<Type>>()))...>;
 };
 
-template <class... Tuples, class Tuple>
-struct for_each_tuple_cat<std::tuple<Tuples...>, std::tuple<Tuple>> {
+template <class... Tuples, class Type>
+struct for_each_tuple_cat<std::tuple<Tuples...>, std::tuple<Type>> {
   using type = std::tuple<decltype(std::tuple_cat(
-      std::declval<Tuples>(), std::declval<std::tuple<Tuple>>()))...>;
+      std::declval<Tuples>(), std::declval<std::tuple<Type>>()))...>;
 };
 
 /// Construct a tuple of tuples that is the result of the concatenation of the
 /// tuples in TupleOfTuples with Tuple.
-template <class TupleOfTuples, class Tuple>
+template <class TupleOfTuples, class Type>
 using for_each_tuple_cat_t =
-    typename for_each_tuple_cat<TupleOfTuples, Tuple>::type;
+    typename for_each_tuple_cat<TupleOfTuples, Type>::type;
 
 template <class InTupleOfTuples, class OutTupleOfTuples>
 struct cartesian_product_impl;
 
+/// \brief Recursive case: The cartesian product is built in a recursive manner.
+/// Each time, each element of the tuple in std::tuple<HeardArgs...> is added to
+/// all the tuples in OutTupleOfTuples. Then the process is repeated with the
+/// next tuple that is present in TailTuples...
+///
+/// \tparam HeadArgs The types of the first tuple in the input tuple of tuples
+/// \tparam TailTuples The remaining tuples in the input tuple of tuples
+/// \tparam OutTupleOfTuples The output tuple of tuples
 template <class... HeadArgs, class... TailTuples, class OutTupleOfTuples>
 struct cartesian_product_impl<
     std::tuple<std::tuple<HeadArgs...>, TailTuples...>, OutTupleOfTuples>
--- a/testing/unit_test/Test_TypeCartesianProduct.cpp
+++ b/testing/unit_test/Test_TypeCartesianProduct.cpp
@@ -1,0 +1,229 @@
+// SPDX-FileCopyrightText: (C) The kokkos-fft development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+#include <gtest/gtest.h>
+#include "KokkosFFT_TypeCartesianProduct.hpp"
+
+// All the tests in this file are compile time tests, so we skip all the tests
+// by GTEST_SKIP(). gtest is used for type parameterization.
+
+namespace {
+using multiple_types = ::testing::Types<int, std::size_t, float, double>;
+
+// Define the types to combine
+using base_real_types = std::tuple<float, double, long double>;
+using base_int_types  = std::tuple<int, std::size_t, bool>;
+
+using multiple_tuple_types = ::testing::Types<base_real_types, base_int_types>;
+
+// Define fixtures
+template <typename T>
+struct CompileTestManipulateTuples : public ::testing::Test {
+  using value_type = T;
+
+  virtual void SetUp() {
+    GTEST_SKIP() << "Skipping all tests for this fixture";
+  }
+};
+
+template <typename T>
+struct CompileTestCartesianProduct : public ::testing::Test {
+  using tuple_type = T;
+
+  virtual void SetUp() {
+    GTEST_SKIP() << "Skipping all tests for this fixture";
+  }
+};
+
+// Tests for appending ValueType or std::tuple<ValueType> to
+// Input tuple type
+template <typename ValueType>
+void test_concat_tuple1D() {
+  using input_tuple_type = std::tuple<std::tuple<double, double>>;
+  using tuple_and_tuple_type =
+      KokkosFFT::Testing::Impl::for_each_tuple_cat_t<input_tuple_type,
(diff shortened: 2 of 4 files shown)


testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,110 @@
...
+namespace Impl {
+
+/// Transform a sequence S to a tuple:
+/// - a std::integer_sequence<T, Ints...> to a
+/// std::tuple<std::integral_constant<T, Ints>...>
+/// - a std::pair<T, U> to a std::tuple<T, U>
+/// - identity otherwise (std::tuple)
+template <class S>
+struct to_tuple {
+  using type = S;
```
science-enthusiast, line 22: Do you want to constrain S to be a scalar using `std::is_scalar`?

testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,110 @@
...
+};
+
+template <class T, T... Ints>
+struct to_tuple<std::integer_sequence<T, Ints...>> {
+  using type = std::tuple<std::integral_constant<T, Ints>...>;
+};
+
+template <class T, class U>
+struct to_tuple<std::pair<T, U>> {
```
science-enthusiast, line 31: Similar to the above comment about the template parameter `S`, do you want to constrain `T` and `U` to be scalars?

testing/src/KokkosFFT_TypeCartesianProduct.hpp
science-enthusiast: Thanks for the clarification.

testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,110 @@
...
+  using type = std::tuple<T, U>;
+};
+
+template <class S>
+using to_tuple_t = typename to_tuple<S>::type;
+
+template <class TupleOfTuples, class Tuple>
+struct for_each_tuple_cat;
```
science-enthusiast, line 39: What does the name `for_each_tuple_cat` signify?

yasahi-hpc: This is coming from the standard [tuple_cat](https://en.cppreference.com/w/cpp/utility/tuple/tuple_cat).
So, basically, it gives a concatenation of all tuples.
Does it make sense?

science-enthusiast: I understand about `std::tuple_cat`. I need to look at further code to understand how it is used. For e.g., why the prefix `for_each`? Also, I need clarification regarding the template parameter name `TupleOfTuples`. Is it a tuple with elements where the elements themselves could be tuples?

science-enthusiast: Your explanation made it clear. Thanks.

yasahi-hpc:
@science-enthusiast 
Do you propose any modifications?

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+#include <gtest/gtest.h>
+#include "KokkosFFT_TypeCartesianProduct.hpp"
+
+// All the tests in this file are compile time tests, so we skip all the tests
+// by GTEST_SKIP(). gtest is used for type parameterization.
+
+namespace {
+using multiple_types = ::testing::Types<int, std::size_t, float, double>;
```
science-enthusiast, line 12: Why is `bool` missing?

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+// Define the types to combine
+using base_real_types = std::tuple<float, double, long double>;
+using base_int_types  = std::tuple<int, std::size_t, bool>;
```
science-enthusiast, line 16: Would you like to order it as `bool`, `int`, `std::size_t`?

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+struct CompileTestMakeCartesianProduct : public ::testing::Test {
+  using tuple_type = T;
+
+  virtual void SetUp() {
+    GTEST_SKIP() << "Skipping all tests for this fixture";
+  }
+};
+
+// Tests for appending ValueType or std::tuple<ValueType> to
+// Input tuple type
```
science-enthusiast, line 49: You could say input tuple having only one tuple as its element.

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+template <typename ValueType>
+void test_concat_tuple1D() {
+  using input_tuple_type = std::tuple<std::tuple<double, double>>;
```
science-enthusiast, line 52: Would you like to parameterize this type also? `test_concat_tuple1D` could have two template parameters: `InputElemType` and `ValueType`.

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+                                                     ValueType>;
+  using reference_tuple_type =
+      std::tuple<std::tuple<double, double, ValueType>>;
+
+  testing::StaticAssertTypeEq<tuple_and_value_type, reference_tuple_type>();
+  testing::StaticAssertTypeEq<tuple_and_tuple_type, reference_tuple_type>();
+}
+
+// Tests for appending ValueType or std::tuple<ValueType> to
+// Input 2D tuple type
```
science-enthusiast, line 67: Similarly you could say input tuple having two tuples as its elements. You could similarly parameterize it with two types `Type1` and `Type2`, say.

testing/unit_test/Test_TypeCartesianProduct.cpp
yasahi-hpc: Could you please detail?
Both inserting ValueType and tuple of ValueType are tested.

testing/unit_test/Test_TypeCartesianProduct.cpp
yasahi-hpc: If I understand, the input tuple having two tuples are tested in test_concat_tuple2D.

testing/unit_test/Test_TypeCartesianProduct.cpp
science-enthusiast: I was wondering why the element type in the inner tuple of `input_tuple_type` are both fixed as `double`. Will it help to make that type also as a template parameter?

testing/unit_test/Test_TypeCartesianProduct.cpp
science-enthusiast: When I read "Input 2D tuple type", it was not clear that it is a tuple with two tuples as elements.

yasahi-hpc: I can rename if you like, say

```C++
test_concat_tuple1D -> test_concat_single_tuple
test_concat_tuple2D -> test_concat_two_tuples
```

testing/unit_test/Test_TypeCartesianProduct.cpp
yasahi-hpc: In this test case, I would like to check that the whatever `ValueType` is actually inserted as a last element. 
So, the fixed `input_tuple_type` is sufficient. 
For a bit more generalization, I can choose different types as an argument of the internal tuple

testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,110 @@
...
+/// std::tuple<std::integral_constant<T, Ints>...>
+/// - a std::pair<T, U> to a std::tuple<T, U>
+/// - identity otherwise (std::tuple)
+template <class S>
+struct to_tuple {
+  using type = S;
+};
+
+template <class T, T... Ints>
+struct to_tuple<std::integer_sequence<T, Ints...>> {
```
science-enthusiast, line 26: Could you show me the place where this specialization is used?

yasahi-hpc: Not used yet. I will add a test for this specialization.
A typical use case is found in DDC
https://github.com/CExA-project/ddc/blob/68f2e99df8a23b769a5a561e3bc8486a4e873a6e/tests/splines/bsplines.cpp#L46-#L50

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+  testing::StaticAssertTypeEq<cartesian_product_type, reference_tuple_type>();
+}
+
+// Tests for getting a cartesian product of types
+// E.g.
+// std::tuple<float, double, long double> would be converted to
+// std::tuple< std::tuple<float>, std::tuple<double>, std::tuple<long double> >
+template <typename TupleType>
+void test_make_cartesian_product_of_tuple1D() {
```
science-enthusiast, line 211: It could be renamed as `test_make_cartesian_product_of_single_tuple`.

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+// E.g.
+// std::tuple<float, double, long double> && std::tuple<float, double, long
+// double> would be converted to std::tuple< std::tuple<float, float>,
+// std::tuple<double, float>, std::tuple<long double, float>,
+//             std::tuple<float, double>, std::tuple<double, double>,
+//             std::tuple<long double, double>, std::tuple<float, long double>,
+//             std::tuple<double, long double>, std::tuple<long double, long
+//             double>>
+template <typename TupleType1, typename TupleType2>
+void test_make_cartesian_product_of_tuple2D() {
```
science-enthusiast, line 234: `test_make_cartesian_product_of_two_tuples`

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+      std::tuple<T0_1, T1_2>, std::tuple<T1_1, T1_2>, std::tuple<T2_1, T1_2>,
+      std::tuple<T0_1, T2_2>, std::tuple<T1_1, T2_2>, std::tuple<T2_1, T2_2>>;
+
+  testing::StaticAssertTypeEq<cartesian_product_type, reference_tuple_type>();
+}
+
+// Tests for getting a cartesian product of types
+// E.g.
+// std::tuple<float, double, long double> && std::tuple<float, double, long
+// double> would be converted to std::tuple< std::tuple<float, float, float>,
```
science-enthusiast, line 163: One more `std::tuple<float, double, long double>` needs to be added.

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+// std::tuple<double, float, float>, std::tuple<long double, float, float>,
+//             ...
+template <typename TupleType1, typename TupleType2, typename TupleType3>
+void test_cartesian_product_of_tuple3D() {
```
science-enthusiast, line 167: `test_cartesian_product_of_three_tuples`

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+// E.g.
+// std::tuple<float, double, long double> && std::tuple<float, double, long
+// double> would be converted to std::tuple< std::tuple<float, float>,
+// std::tuple<double, float>, std::tuple<long double, float>,
+//             std::tuple<float, double>, std::tuple<double, double>,
+//             std::tuple<long double, double>, std::tuple<float, long double>,
+//             std::tuple<double, long double>, std::tuple<long double, long
+//             double>>
+template <typename TupleType1, typename TupleType2>
+void test_cartesian_product_of_tuple2D() {
```
science-enthusiast, line 138: `test_cartesian_product_of_two_tuples`

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+  testing::StaticAssertTypeEq<from_nested_tuple_type,
+                              reference_nested_tuple_type>();
+}
+
+// Tests for getting a cartesian product of types
+// E.g.
+// std::tuple<float, double, long double> would be converted to
+// std::tuple< std::tuple<float>, std::tuple<double>, std::tuple<long double> >
+template <typename TupleType>
+void test_cartesian_product_of_tuple1D() {
```
science-enthusiast, line 115: `test_cartesian_product_of_single_tuple`

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+// Tests for getting a cartesian product of types
+// E.g.
+// std::tuple<float, double, long double> && std::tuple<float, double, long
+// double> would be converted to std::tuple< std::tuple<float, float, float>,
+// std::tuple<double, float, float>, std::tuple<long double, float, float>,
+//             ...
+template <typename TupleType1, typename TupleType2, typename TupleType3>
+void test_make_cartesian_product_of_tuple3D() {
```
science-enthusiast, line 263: `test_make_cartesian_product_of_three_tuples`

testing/unit_test/Test_TypeCartesianProduct.cpp
```diff
@@ -0,0 +1,351 @@
...
+  testing::StaticAssertTypeEq<cartesian_product_type, reference_tuple_type>();
+}
+
+}  // namespace
+
+TYPED_TEST_SUITE(CompileTestManipulateTuples, multiple_types);
+TYPED_TEST_SUITE(CompileTestCartesianProduct, multiple_tuple_types);
+TYPED_TEST_SUITE(CompileTestMakeCartesianProduct, multiple_tuple_types);
+
+TYPED_TEST(CompileTestManipulateTuples, ConcatTuple1D) {
```
science-enthusiast, line 308: The names of the tests need to be changed. They could be `ConcatSingleTuple`, `ConcatTwoTuples`, `ConcatThreeTuples`, `TupleToTypes`, `SingleTuple`, `TwoTuples`, `ThreeTuples`.

science-enthusiast: There seems to be a link between the number of tuples and the dimension of the problem. How about the number of elements in each tuple? What does that signify?

testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+template <class... HeadArgs, class... TailTuples, class OutTupleOfTuples>
+struct cartesian_product_impl<
+    std::tuple<std::tuple<HeadArgs...>, TailTuples...>, OutTupleOfTuples>
+    : cartesian_product_impl<
+          std::tuple<TailTuples...>,
+          decltype(std::tuple_cat(
+              std::declval<for_each_tuple_cat_t<OutTupleOfTuples,
+                                                std::tuple<HeadArgs>>>()...))> {
+};
```
science-enthusiast, line 70: This logic can be described. Something like the following.

The cartesian product is built in a recursive manner. Each time, each element of the tuple in `std::tuple<HeardArgs...>` is added to all the tuples in `OutTupleOfTuples`. Then the process is repeated with the next tuple that is present in `TailTuples...`.

testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,110 @@
...
+struct cartesian_product_impl<std::tuple<>, OutTupleOfTuples> {
+  using type = OutTupleOfTuples;
+};
+
+/// Generate a std::tuple cartesian product from multiple tuple-like structures
+/// (std::tuple, std::integer_sequence and std::pair) Do not rely on the
+/// ordering result.
+template <class... InTuplesLike>
+using cartesian_product_t =
+    typename cartesian_product_impl<std::tuple<to_tuple_t<InTuplesLike>...>,
```
science-enthusiast, line 82: You could add a comment to describe how this is works along with `cartesian_product_impl`.

From the tuple of tuples (`std::tuple<to_tuple_t<InTuplesLike>...>`), the first tuple gets mapped to `std::tuple<HeadArgs...>` in `cartesian_product_impl` and the rest of the tuples to `TailTuples...`.

testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,110 @@
...
+};
+
+template <class OutTupleOfTuples>
+struct cartesian_product_impl<std::tuple<>, OutTupleOfTuples> {
+  using type = OutTupleOfTuples;
+};
+
+/// Generate a std::tuple cartesian product from multiple tuple-like structures
+/// (std::tuple, std::integer_sequence and std::pair) Do not rely on the
+/// ordering result.
```
science-enthusiast, line 79: What does "Do not rely on the ordering result" mean?

testing/unit_test/Test_TypeCartesianProduct.cpp
yasahi-hpc: The number of elements in each tuple is equal to the number of the dimension of cartesian product, i.e., it corresponds to a coordinate.

E.g.
4D Cartesian products -> 4 elements
5D Cartesian products -> 5 elements

science-enthusiast: I don't follow this point. I see that the tests for tuple1D have three elements. The tests for tuple2D have three elements in each tuple and there are two tuples.

yasahi-hpc: tuple1D has 3 elements of tuples with one element
tuple2D has 3x3 elements of tuples with two elements
tuple3D has 3x3x3 elements of tuples with three elements

The number of elements in each tuple corresponds to the dimension

science-enthusiast: I see your point. When you said "The number of elements in each tuple is equal to the number of the dimension of cartesian product", you meant the number of elements in the output tuples. Thus when there are n input tuples, there will be output tuples with n elements each.

yasahi-hpc:
@science-enthusiast 
Thank you for your review.
I have updated accordingly. 

When reviewing, could you please use "Start a review" rather than "Add single comment" to submit the review at one time?
Otherwise, I need to update the code repeatedly when you add a new comment.

science-enthusiast:
> When reviewing, could you please use "Start a review" rather than "Add single comment" to submit the review at one time? Otherwise, I need to update the code repeatedly when you add a new comment.

Noted, Yuuichi :+1:

science-enthusiast:
@yasahi-hpc I see that most of the comments have been addressed. It will be useful to describe what 1D, 2D, 3D etc mean in terms of the input and output tuples. I leave it to you.

yasahi-hpc:
@science-enthusiast 
Thank you for your review. I will merge this. 
I may improve comments when I introduce a documentation for unit-testing tools.

testing/src/KokkosFFT_TypeCartesianProduct.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+template <class T>
+using tuple_to_types_t = typename tuple_to_types<T>::type;
+
+}  // namespace Impl
+
+// API to generate caresian product of input types
+template <class... InTuplesLike>
+using make_cartesian_types =
+    Impl::tuple_to_types_t<Impl::cartesian_product_t<InTuplesLike...>>;
```
science-enthusiast, line 113: You could also add documentation for why `tuple_to_types_t` and `make_cartesian_types` will be needed. It looks like `cartesian_product_t` is enough for practical purposes.

PR #275, opened 2025-05-26
Delegating Plan class constructor

yasahi-hpc:
This PR aims at reusing the multidimensional Plan constructor for the 1D Plan.

### DIFF
--- a/fft/src/KokkosFFT_Plans.hpp
+++ b/fft/src/KokkosFFT_Plans.hpp
@@ -68,6 +68,14 @@ template <typename ExecutionSpace>
 inline constexpr bool is_AllowedSpace_v =
     is_AllowedSpace<ExecutionSpace>::value;
 
+auto inline maybe_null_to_shape(std::optional<std::size_t> n = std::nullopt) {
+  shape_type<1> s = {};
+  if (n) {
+    std::size_t n_tmp = n.value();
+    s                 = shape_type<1>({n_tmp});
+  }
+  return s;
+}
 }  // namespace Impl
 
 /// \brief A class that manages a FFT plan of backend FFT library.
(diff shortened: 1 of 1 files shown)


fft/src/KokkosFFT_Plans.hpp
```diff
@@ -68,6 +68,14 @@ template <typename ExecutionSpace>
 inline constexpr bool is_AllowedSpace_v =
     is_AllowedSpace<ExecutionSpace>::value;
 
+auto inline maybe_null_to_shape(std::optional<std::size_t> n = std::nullopt) {
```
science-enthusiast, line 71: The default value of `std::nullopt` is not needed for a `std::optional` parameter. The `if (n)` check will still work as expected.

science-enthusiast, line 71: Could you describe a situation where the function is called without arguments? Like `maybe_null_to_shape`.

yasahi-hpc: It is used only with an optional argument and is only used for `Plan` creation

yasahi-hpc: So, we can remove the default value

yasahi-hpc: Hi, I have fixed accordingly. Is this good for you?

science-enthusiast (approved):

fft/src/KokkosFFT_Plans.hpp
```diff
@@ -164,60 +172,8 @@ class Plan {
...
-    KOKKOSFFT_THROW_IF(m_is_inplace && m_is_crop_or_pad_needed,
-                       "In-place transform is not supported with reshape. "
-                       "Please use out-of-place transform.");
-
-    KokkosFFT::Impl::setup<ExecutionSpace, float_type>();
-    m_fft_size = KokkosFFT::Impl::create_plan(
-        exec_space, m_plan, in, out, direction, m_axes, s, m_is_inplace);
-  }
+      : Plan(exec_space, in, out, direction, axis_type<DIM>({axis}),
+             Impl::maybe_null_to_shape(n)) {}
```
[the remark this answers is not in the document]
yasahi-hpc: This is a better solution!

PR #276, opened 2025-05-26
Use value_or instead of maybe_null_to_shape

yasahi-hpc:
This PR further simplifies PR #275 by [`value_or`](https://en.cppreference.com/w/cpp/utility/optional/value_or).


tpadioleau (approved):
ok for me

yasahi-hpc:
@tpadioleau Thanks! I will merge this

PR #282, opened 2025-06-08
Suppress disclaimer

yasahi-hpc:
After v0.3.0, we secure the current APIs and thus no longer change APIs without deprecated warnings.


tpadioleau (approved):
Fine with me

yasahi-hpc:
Thank you for approval. I will merge this

PR #295, opened 2025-07-03
Add navier-stokes example

yasahi-hpc:
This PR aims at adding a 3D-NS solver as an advanced example.

- [x] Kokkos version
- [ ] Python version
- [ ] README.md


yasahi-hpc:
> What new features does this example explore compared to "10_HasegawaWakatani" ?

Not much, just more practical.
This can be a good example for distributed FFT

PR #316, opened 2025-09-02
Update rocm Dockerfile to use hipcc 6.2.0

yasahi-hpc:
This PR aims at updating hipcc version in nightly to 6.2.0 which is a new minimum in Kokkos 5.0.


yasahi-hpc:
I have made a look at our Dockerfiles. Since gcc 11.0 is available on ubuntu:22.04, all the images should satisfy new minimum.

PR #300, opened 2025-07-10
Add installation guide with spack

yasahi-hpc:
Resolves #69 

This PR aims at adding an installation guide with spack.

- [x] Add a guide in README.md
- [ ] Add a more comprehensive installation guide in docs
- [ ] Add a CI through spack installation

PR #323, opened 2025-09-11
Add spack installation CI for SYCL backend

### LINKED ISSUE #321: kokkos-fft does not build in a Spack oneapi stack
log from Spack CI (https://gitlab.spack.io/spack/spack-packages/-/jobs/18060719): [oneapi.log](https://github.com/user-attachments/files/22254260/oneapi.log)

yasahi-hpc:
Resolves #321

This PR aims at adding a spack installation CI for sycl environment.
Documentation will be done in https://github.com/kokkos/kokkos-fft/pull/300

### DIFF
--- a/docker/intel/Dockerfile
+++ b/docker/intel/Dockerfile
@@ -13,21 +13,19 @@ RUN apt-get update && apt-get install -y \
         git \
         ccache \
         libfftw3-dev \
-        python3 \
-        gfortran \
         $ADDITIONAL_PACKAGES \
         && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-ENV FFTWDIR=/usr \
-    CMAKE_DIR=/opt/cmake
+ENV FFTWDIR "/usr"
 
 RUN git config --global --add safe.directory "*"
 
 # Install newer CMake manually
 ARG CMAKE_VERSION=3.25.2
 
+ENV CMAKE_DIR=/opt/cmake
 RUN CMAKE_URL=https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION} && \
     CMAKE_SCRIPT=cmake-${CMAKE_VERSION}-linux-x86_64.sh && \
     CMAKE_SHA256=cmake-${CMAKE_VERSION}-SHA-256.txt && \
@@ -48,3 +46,4 @@ ENV PATH=${CMAKE_DIR}/bin:$PATH
 
 WORKDIR /work
 CMD ["bash"]
+


yasahi-hpc:
@tpadioleau 
Thank you for the fix. 
Do you know why we need to specify the compiler?

tpadioleau:
> @tpadioleau Thank you for the fix. Do you know why we need to specify the compiler?

Actually the fix is not really a fix, the docker image contains two oneapi installations:
- the one from the image
- the one from Spack

When Spack installed version 2024, there was a mix of the two: using include files from one installation and linking to the other installation resulting in a ABI incompatibility.

I failed to tell Spack to reuse the installation from the image. We need to figure out to proceed.

tpadioleau:
> Do you know why we need to specify the compiler?

I think Kokkos only imposes the compiler for the CUDA backend when using the wrapper. For the rest of the backends, i think it is the responsability of the user to use a compatible compiler.

yasahi-hpc:
> > Do you know why we need to specify the compiler?
> 
> I think Kokkos only imposes the compiler for the CUDA backend when using the wrapper. For the rest of the backends, i think it is the responsability of the user to use a compatible compiler.

Makes sense. I thought spack manages the dependency to the compiler as well

.github/workflows/spack-test.yaml
```diff
@@ -42,5 +43,40 @@ jobs:
...
+          bash -c "apt update && \
+            apt install --yes --no-install-recommends file bzip2 ca-certificates g++ gcc gfortran git gzip lsb-release patch python3 tar unzip xz-utils zstd && \
+            git clone --depth=1 --branch=v1.0.2 https://github.com/spack/spack.git && \
+            . spack/share/spack/setup-env.sh && \
+            spack install intel-oneapi-compilers@2025 && \
+            spack spec ${{ matrix.backend.spec }} && \
+            spack install ${{ matrix.backend.spec }} && \
+            spack load cmake intel-oneapi-compilers kokkos kokkos-fft && \
+            cd /work/install_test/as_library && \
+            cmake -D CMAKE_CXX_COMPILER=${{ matrix.backend.compiler }} -B build && \
```
yasahi-hpc, line 81: Does it make sense to create a new image for this?

yasahi-hpc: Sure. We cannot rely on gcc container which is almost identical to the official image of Ubuntu

yasahi-hpc (commented):
LGTM

docker/intel/Dockerfile
```diff
@@ -13,19 +13,21 @@ RUN apt-get update && apt-get install -y \
         git \
         ccache \
         libfftw3-dev \
+        python3 \
+        gfortran \
```
tpadioleau, line 17: Do we still need it ?

yasahi-hpc: No, we can revert this change. 
Clean up the docker `ENV` should be made in another PR

yasahi-hpc: Does this resolve #321 ?

tpadioleau: > Does this resolve #321 ?

I think yes, we can conclude that kokkos-fft is fine, the problem is on the Spack side.

yasahi-hpc: > Ok so we remove all changes in this Dockerfile ?

Yes, please

tpadioleau:
> Thanks for the update. Can I merge this?

Yes I think it is ready.

PR #337, opened 2025-10-17
Extract the core logic of get_map_axes helper

yasahi-hpc:
This PR aims at extracting the core logic of `get_map_axes` which does not rely on `View`. 
For distributed FFT, we do not have an access to the global `View` which does not necessarily exist.
The extracted function can compute the mapping from the meta data of global `View` rather than `View` itself.

### DIFF
--- a/common/src/KokkosFFT_transpose.hpp
+++ b/common/src/KokkosFFT_transpose.hpp
@@ -52,7 +52,7 @@ auto get_map_axes(const std::array<IntType, FFT_DIM>& axes) {
     }
 
     // Then stack remaining axes
-    for (iType i = 0; i < rank; i++) {
+    for (IntType i = 0; i < rank; i++) {
       if (!is_found(non_negative_axes, i)) {
         map.push_back(i);
       }


common/src/KokkosFFT_transpose.hpp
```diff
@@ -47,23 +51,31 @@ auto get_map_axes(const ViewType& view, axis_type<DIM> axes) {
...
+  for (iType i = 0; i < rank; i++) {
     array_map_inv.at(i) = get_index(array_map, i);
   }
 
-  return std::tuple<full_axis_type, full_axis_type>({array_map, array_map_inv});
+  return std::make_tuple(array_map, array_map_inv);
+}
+
+template <typename ViewType, std::size_t DIM>
+auto get_map_axes(const ViewType& view, axis_type<DIM> axes) {
```
science-enthusiast: Is `axis_type<DIM>` always of the form `std::array<iType, DIM>`?

Also, the name `DIM` that is used here has a different meaning to the `DIM` in the other function (it is `ViewType::rank()` in the other case).

yasahi-hpc:
@science-enthusiast 

Thanks! I will merge this

PR #342, opened 2025-10-21
Make compute_strides function a public function

yasahi-hpc:
This PR aims at moving `compute_strides` under `KokkosFFT::Impl` namespace. This function can be used not only for rocFFT but also for [cuFFTMp](https://github.com/yasahi-hpc/distributed-FFT-for-kokkos/blob/781a9bcc14b17015716a47e5c04ad8ebafad497d/distributed/src/cuFFTMp/KokkosFFT_Distributed_CufftMp_Plan.hpp#L68-L84) with small differences.

### DIFF
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -391,11 +391,16 @@ auto convert_base_int_type(const ContainerType& src) {
   }
 }
 
-// \brief Helper to compute strides from extents
+// \brief Helper to compute strides from extents.
+// The extents are computed from a LayoutRight View.
+// The computed strides can be considered as view strides
+// in the reversed order.
+//
 // Examples:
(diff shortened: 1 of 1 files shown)


common/src/KokkosFFT_utils.hpp
```diff
@@ -391,6 +391,37 @@ auto convert_base_int_type(const ContainerType& src) {
   }
 }
 
+// \brief Helper to compute strides from extents
+// Examples:
+// (n0, n1, n2) -> (1, n2, n1*n2)
+// (n0, n1) -> (1, n1)
+// (n0) -> (1)
```
PaulGannay: I understood the old 
```
  // (n0, n1, n2) -> (1, n0, n0*n1)
  // (n0, n1) -> (1, n0)
  // (n0) -> (1)
```
but I can't make sense of this example.
Could you explain what kind of stride you are computing?

PR #343, opened 2025-10-21
Introduce helper function to wrap std::reverse

yasahi-hpc:
This PR aims at introducing a thin wrapper for `std::reverse` which is frequently used in the code


yasahi-hpc:
@PaulGannay 

Thanks! I will merge this

PR #345, opened 2025-10-26
Copy if map is identical in transpose helper

yasahi-hpc:
This PR aims at allowing transpose function to perform `deep_copy` if transpose is not needed.

- [x] Using execution space instance appropriately in transpose unit-tests
- [x] Skip some case when map is identical. I found that the test will result in test failures in `fftshift` tests in `Release` build on `Cuda` backend. This does not happen in CPU backends or Cuda backend with `Debug` build. This happens only with 4D cases.

### DIFF
--- a/common/unit_test/Test_Transpose.cpp
+++ b/common/unit_test/Test_Transpose.cpp
@@ -60,11 +60,6 @@ void make_transposed(const ViewType1& x, const ViewType2& xT,
   static_assert(ViewType1::rank() == DIM && ViewType2::rank() == DIM,
                 "make_transposed: Rank of Views must be equal to Rank of "
                 "transpose axes.");
-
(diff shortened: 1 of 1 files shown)


PaulGannay (approved):
The changes in `KokkosFFT_transpose.hpp` looks correct, but the test file is so big that it is becoming hard to review, it would be best if you could find a way to factorise some of the code to make the changes easier to track.
It is orthogonal with the work done in this PR though.

tretre91 (approved):
looks good to me

yasahi-hpc:
Thanks @PaulGannay @tretre91 
I will refactor first in another PR
As you said, tests can be simplified

yasahi-hpc:
@PaulGannay I opened #348 focusing on the execution space instance. Can I have your review please.

PR #351, opened 2025-10-31
Rename iType to IndexType to align with mdspan convention

yasahi-hpc:
This PR aims at improving the naming convention (see [mdspan convention](https://en.cppreference.com/w/cpp/container/mdspan.html)). 
We now use `IndexType` to represent the int type corresponding for `indices`. 
We still keep `IntType` for integers which are not used for indices, such as `axes`

### DIFF
--- a/testing/src/KokkosFFT_CountErrors.hpp
+++ b/testing/src/KokkosFFT_CountErrors.hpp
@@ -51,10 +51,10 @@ Kokkos::Iterate get_iteration_order(const ViewType& view) {
 /// \tparam ComparisonOp The binary operation to apply to the elements of the
 /// views.
 /// \tparam Layout The layout type of the Kokkos views.
-/// \tparam iType The integer type used for indexing the view elements.
+/// \tparam IndexType The integer type used for indexing the view elements.
 template <KokkosExecutionSpace ExecutionSpace, KokkosView AViewType,
           KokkosView BViewType, typename ComparisonOp, KokkosLayout Layout,
-          typename iType>
+          typename IndexType>
 struct ViewErrors {
  private:
   // Since MDRangePolicy is not available for 7D and 8D views, we need to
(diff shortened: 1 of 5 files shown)


testing/src/KokkosFFT_CountErrors.hpp
```diff
@@ -125,7 +125,7 @@ struct ViewErrors {
   template <std::size_t... Idx>
   struct CountErrors<std::index_sequence<Idx...>> {
     template <std::size_t I>
-    using IndicesType = iType;
+    using IndicesType = IndexType;
```
PaulGannay: Could this be deleted? It looks redundant.

yasahi-hpc: Can I do that in another PR?

yasahi-hpc: Sure
Let me do that in another PR
I try to make PRs as small as possible

yasahi-hpc: Actually, I could not find a workaround other than this one.
Do you have any suggestions?

This does not compile

```C++
  template <std::size_t... Idx>
  struct CountErrors<std::index_sequence<Idx...>> {
    //template <std::size_t I>
    //using IndicesType = IndexType;
    
    AViewType m_a;
    BViewType m_b;

    ComparisonOp m_op;

    /// \brief Constructs the error counter and performs the error computation.
    ///
    /// \param[in] a First Kokkos view containing data to compare.
    /// \param[in] b Second Kokkos view containing data to compare against.
    /// \param[in] p The binary operation used for comparison
    CountErrors(const AViewType& a, const BViewType& b, ComparisonOp op)
        : m_a(a), m_b(b), m_op(op) {}

    /// \brief Operator called by Kokkos to perform the comparison of each
    /// element.
    ///
    /// \param[in] indices The indices of the element in the views up to 6D.
    /// \param[in,out] err  The error counter incremented if a mismatch is
    /// detected.
    KOKKOS_INLINE_FUNCTION
    void operator()(const IndexType... indices, std::size_t& err) const {
```

PR #352, opened 2025-11-02
Testing transpose with bound checks

yasahi-hpc:
This PR aims at improving tests for transpose helper. 

- [x] When `bounds_check` is true, testing the cases with different input and output extents.


common/unit_test/Test_Transpose.cpp
```diff
@@ -376,18 +376,25 @@ void test_transpose_1d_1dview(bool bounds_check) {
   // When transpose is not necessary, we should not call transpose method
   using View1DLayout1type = Kokkos::View<double*, LayoutType1, execution_space>;
   using View1DLayout2type = Kokkos::View<double*, LayoutType2, execution_space>;
-  const int len           = 30;
+  const int len = 30, perturbation = bounds_check ? -2 : 0;
```
PaulGannay:
```suggestion
  const int len           = 30;
  const int perturbation = bounds_check ? -2 : 0;
```
(You will have to re-execute the clang-format, assignment won't be aligned).

PR #354, opened 2025-11-07
fix: int type in transpose

yasahi-hpc:
This PR fixes the base int type of map in transpose helper.
Previously, it only worked for `std::array<int, DIM>`.
Now, it works also for `std::array<std::size_t, DIM>`

### DIFF
--- a/common/src/KokkosFFT_transpose.hpp
+++ b/common/src/KokkosFFT_transpose.hpp
@@ -145,10 +145,12 @@ struct BoundsCheck {
 /// \tparam ExecutionSpace The type of Kokkos execution space.
 /// \tparam InViewType The input view type
 /// \tparam OutViewType The output view type
-/// \tparam IndexType The index type used for the map
+/// \tparam IndexType The index type used for the loop
+/// \tparam IntType The int type used for the map
 /// \tparam ArgBoundsCheck The bounds check type (default is Off)
 template <typename ExecutionSpace, typename InViewType, typename OutViewType,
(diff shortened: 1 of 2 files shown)


yasahi-hpc:
> Since the map is holding index value used to access views, why do you need a new type in addition to `IndexType`?

Map does not hold the index value used to access views.
It holds the permutation array that maps src and dst index arrays.

```C++
m_out(src_idx[m_map[Is]]...) = m_in(src_idx[Is]...);
```

As you said, `src_idx` must be in `IndexType`, but `int` type of the `m_map` can be different.

PaulGannay:
> Actually, I am considering that. Although APIs allow `std::array<int, DIM>` to specify axes, they should be converted into `std::array<std::size_t, DIM>` internally
> Since it requires more changes, I would like to do that in another PR

I would close this PR without merge then.
But if you think that it is necessary, in preparation of other work, I haven't seen anything wrong either so I guess that you can merge it.

yasahi-hpc:
> > Actually, I am considering that. Although APIs allow `std::array<int, DIM>` to specify axes, they should be converted into `std::array<std::size_t, DIM>` internally
> > Since it requires more changes, I would like to do that in another PR
> 
> I would close this PR without merge then. But if you think that it is necessary, in preparation of other work, I haven't seen anything wrong either so I guess that you can merge it.

It is useful for distributed implementation details, wherein all the axes are already converted into `std::array<std::size_t, DIM>`. If you prefer, I can change this function to work on `std::array<std::size_t, DIM>` and convert maps into `std::array<std::size_t, DIM>`

PR #355, opened 2025-11-07
Improve iteration pattern in transpose

yasahi-hpc:
This PR aims at choosing optimal iteration pattern for MDRange based on the `array_layout`.
Since it is under `Impl` namespace, I cannot directly use the Kokkos implementation.
https://github.com/kokkos/kokkos/blob/cb4fbfd8487ef2f07b2c845a70802bdd71a1c3de/core/src/Kokkos_Layout.hpp#L222-L250

### DIFF
--- a/common/src/KokkosFFT_utils.hpp
+++ b/common/src/KokkosFFT_utils.hpp
@@ -454,10 +454,9 @@ auto reversed(Container&& c) {
 // to avoid dependency on Kokkos implementation details
 template <typename... Layout>
 struct layout_iterate_type_selector {
-  static const Kokkos::Iterate outer_iteration_pattern =
-      Kokkos::Iterate::Default;
-  static const Kokkos::Iterate inner_iteration_pattern =
-      Kokkos::Iterate::Default;
+  static_assert(true,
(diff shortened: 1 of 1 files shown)


common/src/KokkosFFT_transpose.hpp
```diff
@@ -172,9 +172,14 @@ struct Transpose {
           Kokkos::RangePolicy<ExecutionSpace, Kokkos::IndexType<IndexType>>;
       return range_policy_type(space, 0, x.extent(0));
     } else {
+      using LayoutType = typename InViewType::array_layout;
```
science-enthusiast: `array_layout` is to be deprecated soon in Kokkos Views: https://github.com/kokkos/kokkos/pull/7401

yasahi-hpc: I am not sure when to use the new name
The `array_layout` is the only one found in [documentation](https://kokkos.org/kokkos-core-wiki/API/core/view/view.html#_CPPv4N4View12array_layoutE)

I will probably do that later

science-enthusiast (approved):
Looks good to me.

PR #433, opened 2026-04-11
Add rfft/irfft example

yasahi-hpc:
This PR aims at adding a minimum working example in API reference. 

- [x] Add docs_rfft.cpp which includes a minimum working example that is compiled in CI.
- [x] Add docs_irfft.cpp which includes a minimum working example that is compiled in CI.
- [x] Embed this in API reference with literal includes.

### DIFF
--- a/docs/api/real/irfft.rst
+++ b/docs/api/real/irfft.rst
@@ -9,7 +9,7 @@ KokkosFFT::irfft
 
 .. note::
 
-   The input must be a complex-valued view, and the output must be a real-valued view. The input size along the transform axis is ``n/2 + 1``, where ``n`` is the output size along that axis.
+   The input must be a complex-valued view, and the output must be a real-valued view. The input length along the transform axis is ``n/2 + 1``, where ``n`` is the output length along that axis. If this condition is not met, the `std::runtime_error` exception will be thrown.
 
 Examples
 ========
--- a/docs/api/real/irfft.rst
+++ b/docs/api/real/irfft.rst
@@ -6,3 +6,21 @@ KokkosFFT::irfft
 ----------------
 
 .. doxygenfunction:: KokkosFFT::irfft(const ExecutionSpace& exec_space, const InViewType& in, const OutViewType& out, KokkosFFT::Normalization, int axis, std::optional<std::size_t> n)
+
+.. note::
+
+   The input must be a complex-valued view, and the output must be a real-valued view. The input size along the transform axis is ``n/2 + 1``, where ``n`` is the output size along that axis.
+
+Examples
+========
+
+.. literalinclude:: ../../../examples/docs/real/docs_irfft.cpp
+  :language: c++
+  :linenos:
+  :lines: 5-
+
+Expected output:
+
+.. code::
+
+  1 2 3 4
--- a/docs/api/real/rfft.rst
+++ b/docs/api/real/rfft.rst
@@ -9,7 +9,7 @@ KokkosFFT::rfft
 
 .. note::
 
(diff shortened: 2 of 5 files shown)


Adrien-Tab (approved):

yasahi-hpc:
Thanks @Adrien-Tab 
I will merge this
I just found a small typo which has been fixed

examples/docs/CMakeLists.txt
```diff
@@ -19,3 +20,10 @@ target_link_libraries(docs_fftn PUBLIC KokkosFFT::fft)
 
 add_executable(docs_ifftn standard/docs_ifftn.cpp)
```
daxmawal, line 21: It should be ifft -> irfft ?

docs/api/real/rfft.rst
```diff
@@ -5,4 +5,22 @@
 KokkosFFT::rfft
 ---------------
 
-.. doxygenfunction:: KokkosFFT::rfft(const ExecutionSpace& exec_space, const InViewType& in, const OutViewType& out, KokkosFFT::Normalization, int axis, std::optional<std::size_t> n)
\ No newline at end of file
+.. doxygenfunction:: KokkosFFT::rfft(const ExecutionSpace& exec_space, const InViewType& in, const OutViewType& out, KokkosFFT::Normalization, int axis, std::optional<std::size_t> n)
+
+.. note::
+
+   The input must be a real-valued view, and the output must be a complex-valued view. The output length along the transform axis is ``n/2 + 1``, where ``n`` is the input length along that axis. If this condition is not met, the `std::runtime_error` exception will be thrown.
```
daxmawal, line 12: Sorry I'm late, 

I may be missing something, but when n is provided, the implementation seems to derive a modified shape and call crop_or_pad() if needed.

https://github.com/kokkos/kokkos-fft/blob/3ab2d455b6153cfe00971be37b107cb9fd0195cc/fft/src/KokkosFFT_Plans.hpp#L245

docs/api/real/irfft.rst
```diff
@@ -5,4 +5,22 @@
 KokkosFFT::irfft
 ----------------
 
-.. doxygenfunction:: KokkosFFT::irfft(const ExecutionSpace& exec_space, const InViewType& in, const OutViewType& out, KokkosFFT::Normalization, int axis, std::optional<std::size_t> n)
\ No newline at end of file
+.. doxygenfunction:: KokkosFFT::irfft(const ExecutionSpace& exec_space, const InViewType& in, const OutViewType& out, KokkosFFT::Normalization, int axis, std::optional<std::size_t> n)
+
+.. note::
+
+   The input must be a complex-valued view, and the output must be a real-valued view. The input length along the transform axis is ``n/2 + 1``, where ``n`` is the output length along that axis. If this condition is not met, the `std::runtime_error` exception will be thrown.
```
daxmawal, line 12: Maybe a crop_or_pad() is also apply here, on the input view ?

yasahi-hpc: Here, `n` just refers the length of transform and is unrelated to the optional parameter of `n`.
I agree it is confusing, but it is common to use `n` to represent the length

docs/api/real/rfft.rst
yasahi-hpc: That is perfectly fine
Thanks again for your review
See my comments above

examples/docs/CMakeLists.txt
yasahi-hpc: New examples are added at the bottom. So, this should be fine

PR #435, opened 2026-04-14
Refactor padding helper

yasahi-hpc:
- [x] Rename header from `KokkosFFT_padding.hpp` to `KokkosFFT_Padding.hpp`
- [x] Refactor headers in `KokkosFFT_Padding.hpp`
- [x] Remove `is_crop_or_pad_needed` which can be simply achieved by `in_extents != extents`
- [x] Remove tests for `is_crop_or_pad_needed`
- [x] Add docstings to `crop_or_pad` and `crop_or_pad_impl`

### DIFF
--- a/common/src/KokkosFFT_Padding.hpp
+++ b/common/src/KokkosFFT_Padding.hpp
@@ -12,20 +12,30 @@
 #include <Kokkos_Core.hpp>
 #include "KokkosFFT_common_types.hpp"
 #include "KokkosFFT_asserts.hpp"
+#include "KokkosFFT_traits.hpp"
 
 namespace KokkosFFT {
 namespace Impl {
-/// \brief Partially copy the input view to the output view to pad or crop the
-/// input view to the output view.
-/// \tparam ExecutionSpace The type of Kokkos execution space.
-/// \tparam InViewType The input view type
-/// \tparam OutViewType The output view type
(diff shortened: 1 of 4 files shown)


common/src/KokkosFFT_Padding.hpp
```diff
@@ -0,0 +1,84 @@
...
+template <typename ExecutionSpace, typename SrcViewType, typename DstViewType,
+          std::size_t... Is>
+void crop_or_pad_impl(const ExecutionSpace& exec_space, const SrcViewType& src,
+                      const DstViewType& dst, std::index_sequence<Is...>) {
+  constexpr std::size_t rank = SrcViewType::rank();
+  using extents_type         = std::array<std::size_t, rank>;
+
+  extents_type extents{};
+  for (std::size_t i = 0; i < rank; i++) {
+    extents.at(i) = std::min(src.extent(i), dst.extent(i));
```
daxmawal: The loop is bounded by `i < rank`, using `extents[i]` instead of `extents.at(i)` would avoid an extra bounds check.

yasahi-hpc: That is true

yasahi-hpc:
Thanks for your review @daxmawal 
I have updated accordingly

Can I have another review

yasahi-hpc:
I will merge this
Thanks for your review @daxmawal

PR #444, opened 2026-04-17
Add MD operation helper

yasahi-hpc:
This PR aims to extract a helper function to create a Range/MDRange policy for multidimensional operations.

- [x] Introduce a new header `KokkosFFT_MDOperations.hpp`
- [x] Introduce `get_mdpolicy` function with tests
- [x] Use `get_mdpolicy` in `Transpose` and `Roll` struct

### DIFF
--- a/common/src/KokkosFFT_Helpers.hpp
+++ b/common/src/KokkosFFT_Helpers.hpp
@@ -9,6 +9,7 @@
 #include "KokkosFFT_common_types.hpp"
 #include "KokkosFFT_Traits.hpp"
 #include "KokkosFFT_Layout.hpp"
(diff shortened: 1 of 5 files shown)


Adrien-Tab:
It's look similar to what is done here https://github.com/kokkos/kokkos/blob/develop/core/unit_test/view/TestBasicView.hpp

yasahi-hpc:
That is true. But these are not APIs

Adrien-Tab (approved):
LGTM

yasahi-hpc:
Thank you for approval @Adrien-Tab 
I will merge this

PR #445, opened 2026-04-20
Rename KokkosFFT_transpose.hpp to KokkosFFT_Transpose.hpp

yasahi-hpc:
- [x] Rename header from `KokkosFFT_transpose.hpp` to `KokkosFFT_Transpose.hpp`
- [x] Move `is_transpose_needed` from `KokkosFFT_utils.hpp` to `KokkosFFT_Transpose.hpp`
- [x] Fix: missing includes of `KokkosFFT_Transpose.hpp`
- [x] Quick return when out-of-bound access happens

### DIFF
--- a/common/src/KokkosFFT_Transpose.hpp
+++ b/common/src/KokkosFFT_Transpose.hpp
@@ -5,8 +5,12 @@
 #ifndef KOKKOSFFT_TRANSPOSE_HPP
 #define KOKKOSFFT_TRANSPOSE_HPP
 
+#include <array>
+#include <limits>
 #include <numeric>
 #include <tuple>
+#include <type_traits>
+#include <Kokkos_Core.hpp>
 #include "KokkosFFT_common_types.hpp"
 #include "KokkosFFT_utils.hpp"
 #include "KokkosFFT_Padding.hpp"
@@ -19,6 +23,23 @@ struct BoundsCheck {
   struct On {};
   struct Off {};
 };
+
+/// \brief Check if transpose is needed or not
+/// If a map is contiguous and in ascending order (e.g. {0, 1, 2}),
+/// we do not need transpose
+/// \tparam IndexType The integer type used for map
+/// \tparam DIM The dimensionality of the axes
+///
+/// \param[in] map The map used for permutation
+/// \return true if transpose is needed, false otherwise
+template <typename IndexType, std::size_t DIM>
+bool is_transpose_needed(const std::array<IndexType, DIM>& map) {
+  static_assert(std::is_integral_v<IndexType>,
+                "is_transpose_needed: IndexType must be an integral type.");
+  std::array<IndexType, DIM> contiguous_map;
+  std::iota(contiguous_map.begin(), contiguous_map.end(), 0);
+  return map != contiguous_map;
+}
 
 /// \brief Transpose functor for out-of-place transpose operations.
 /// This struct implements a functor that applies a transpose on a Kokkos view.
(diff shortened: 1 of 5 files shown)


common/unit_test/Test_Transpose.cpp
```diff
@@ -94,6 +103,27 @@ void make_transposed(const ViewType1& x, const ViewType2& xT,
...
+  std::array<IndexType, 1> map1D{0};
+  EXPECT_FALSE(KokkosFFT::Impl::is_transpose_needed(map1D));
+
+  std::array<IndexType, 2> map2D{0, 1}, map2D_axis0{1, 0};
+  EXPECT_FALSE(KokkosFFT::Impl::is_transpose_needed(map2D));
+  EXPECT_TRUE(KokkosFFT::Impl::is_transpose_needed(map2D_axis0));
+
+  std::array<IndexType, 3> map3D{0, 1, 2}, map3D_021{0, 2, 1},
+      map3D_102{1, 0, 2}, map3D_120{1, 2, 0}, map3D_201{2, 0, 1},
+      map3D_210{2, 1, 0};
```
PaulGannay, line 117:
```suggestion
  std::array<IndexType, 3> map3D     = {0, 1, 2};
  std::array<IndexType, 3> map3D_021 = {0, 2, 1};
  std::array<IndexType, 3> map3D_102 = {1, 0, 2};
  std::array<IndexType, 3> map3D_120 = {1, 2, 0};
  std::array<IndexType, 3> map3D_201 = {2, 0, 1};
  std::array<IndexType, 3> map3D_210 = {2, 1, 0};
```
I found the old format easier to read, why change it?

common/src/KokkosFFT_Transpose.hpp
```diff
@@ -94,18 +115,14 @@ struct Transpose {
         IndexType src_idx[], std::index_sequence<Is...>) const {
       if constexpr (std::is_same_v<ArgBoundsCheck, BoundsCheck::On>) {
         // Bounds check
-        bool in_bounds = true;
         for (std::size_t i = 0; i < InViewType::rank(); ++i) {
```
science-enthusiast, line 118: Maybe, a while loop is more suitable than a for loop, when an exit could happen inbetween.

PaulGannay (commented):
LGTM

common/unit_test/Test_Transpose.cpp
PaulGannay: This may be better but I guess it doesn't pass clang-format?
```C++
  std::array<IndexType, 3> map3D{0, 1, 2},
                           map3D_021{0, 2, 1},
                           map3D_102{1, 0, 2},
                           map3D_120{1, 2, 0},
                           map3D_201{2, 0, 1},
                           map3D_210{2, 1, 0};
```

yasahi-hpc: Unfortunately, it doesn't pass clang-format. The previous one may be easier to read, but it uses more lines.

common/src/KokkosFFT_Transpose.hpp
yasahi-hpc: I often prefer `for` loop when the loop bound is clear. I use `while` loop if the condition is not described by a loop structure, say, https://github.com/kokkos/kokkos-kernels/blob/4f4eee62fecb60e1cb158c5192a14c993c780886/batched/dense/impl/KokkosBatched_Getrf_Serial_Internal.hpp#L174-#L297

common/src/KokkosFFT_Transpose.hpp
```diff
@@ -5,8 +5,12 @@
 #ifndef KOKKOSFFT_TRANSPOSE_HPP
 #define KOKKOSFFT_TRANSPOSE_HPP
```
daxmawal, line 7: You may also want to include utility, since std::index_sequence is declared there.

yasahi-hpc: Nice catch.
I will fix that

I will also refactor the detailed copilot instructions as well for more comprehensive include check

yasahi-hpc:
Thank you for your review @PaulGannay @science-enthusiast @daxmawal 
I have further updated the include files
May I have another review

science-enthusiast (approved):
Looks good to me.

yasahi-hpc:
Thanks @science-enthusiast 
I will merge this

PR #446, opened 2026-04-22
Add md operator helper

yasahi-hpc:
This PR aims at implementing a broadcasting operator to work on a `View`.
It applies operator `op` to all the elements of `x`

```C++
x = op(x); 
```

- [x] Add a MD operator wrapper with `MDRange` to do `x = op(x)`
- [x] Use the operator in `normalize_impl`, `hfft` and `ihfft`
- [x] Remove badly implemented `conjugate` helper used in `hfft` and `ihfft`

### DIFF
--- a/common/src/KokkosFFT_UnaryOps.hpp
+++ b/common/src/KokkosFFT_UnaryOps.hpp
@@ -21,7 +21,11 @@ struct Conjugate {
   /// input value unchanged.
   template <typename T>
   KOKKOS_INLINE_FUNCTION auto operator()(const T& x) const {
-    return is_complex_v<T> ? Kokkos::conj(x) : x;
+    if constexpr (is_complex_v<T>) {
+      return Kokkos::conj(x);
+    } else {
+      return x;
+    }
   }
 };
 
--- a/common/src/KokkosFFT_UnaryOps.hpp
+++ b/common/src/KokkosFFT_UnaryOps.hpp
@@ -1,0 +1,126 @@
+// SPDX-FileCopyrightText: (C) The Kokkos-FFT development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+#ifndef KOKKOSFFT_UNARYOPS_HPP
+#define KOKKOSFFT_UNARYOPS_HPP
+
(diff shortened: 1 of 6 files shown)


common/src/KokkosFFT_UnaryOps.hpp
```diff
@@ -0,0 +1,128 @@
...
+  /// complex, otherwise returns the input value unchanged.
+  /// \tparam T The type of the input value.
+  /// \param[in] x The input value.
+  /// \return The conjugate of the input value if it is complex, otherwise the
+  /// input value unchanged.
+  template <typename T>
+  KOKKOS_INLINE_FUNCTION auto operator()(const T& x) const {
+    return is_complex_v<T> ? Kokkos::conj(x) : x;
+  }
+};
```
tretre91, line 30: Just to be sure, is it intended that this function will always return `Kokkos::complex` even when `T` is not a complex type ? (since `Kokkos::conj` always return a complex, the other branch will always return `Kokkos::complex(x)`)

yasahi-hpc: Good catch
That is a mistake

tretre91 (approved):

common/src/KokkosFFT_UnaryOps.hpp
yasahi-hpc: Strictly speaking, should we do something like this?
Can we avoid copy for non-complex rvalue this way?

```
KOKKOS_INLINE_FUNCTION decltype(auto) operator()(T&& x) const {
  if constexpr (is_complex_v<std::decay_t<T>>) {
    return Kokkos::conj(x);
  } else {
    return std::forward<T>(x);
  }
}
```

tretre91: I'd say if you only expect scalars then it's not worth the trouble, but if the function will eventually be used with something other than a scalar or a complex then yes it could be useful to do it this way

yasahi-hpc: Thank you for the comments.
As named, it is mostly used for complex values, so it may not matter

yasahi-hpc:
Thank you for review and approval @tretre91 
I will merge this

PR #448, opened 2026-04-24
fix: transpose tests

yasahi-hpc:
For some reason, transpose tests triggers failures on NVIDIA backend. This PR aims at resolving the failure by isolating transpose tests.

- [x] Separate transpose tests
- [x] Minor cleanup of Test_Main.cpp 
- [x] Complete refactor of Test_Transpose.cpp to generalize the test functions. Also, a helper function `make_transposed` to create a reference has also been refactored to avoid out-of-rank View `extent` access.

### DIFF
--- a/common/unit_test/CMakeLists.txt
+++ b/common/unit_test/CMakeLists.txt
@@ -10,17 +10,20 @@ add_executable(
   Test_Layout.cpp
   Test_Normalization.cpp
   Test_Mapping.cpp
-  Test_MDOperations.cpp
-  Test_Transpose.cpp
   Test_Extents.cpp
+  Test_MDOperations.cpp
   Test_Padding.cpp
   Test_Helpers.cpp
 )
 
 target_compile_features(unit-tests-kokkos-fft-common PUBLIC cxx_std_17)
-
 target_link_libraries(unit-tests-kokkos-fft-common PUBLIC common GTest::gtest)
 
 # Enable GoogleTest
 include(GoogleTest)
 gtest_discover_tests(unit-tests-kokkos-fft-common PROPERTIES DISCOVERY_TIMEOUT 600 DISCOVERY_MODE PRE_TEST)
+
+add_executable(unit-tests-kokkos-transpose Test_Main.cpp Test_Transpose.cpp)
+target_compile_features(unit-tests-kokkos-transpose PUBLIC cxx_std_17)
+target_link_libraries(unit-tests-kokkos-transpose PUBLIC KokkosFFT::fft GTest::gtest)
+gtest_discover_tests(unit-tests-kokkos-transpose PROPERTIES DISCOVERY_TIMEOUT 600 DISCOVERY_MODE PRE_TEST)
--- a/common/unit_test/CMakeLists.txt
+++ b/common/unit_test/CMakeLists.txt
@@ -9,9 +9,9 @@ add_executable(
   Test_Traits.cpp
   Test_Layout.cpp
   Test_Normalization.cpp
+  Test_MDOperations.cpp
   Test_Mapping.cpp
   Test_Extents.cpp
-  Test_MDOperations.cpp
   Test_Padding.cpp
   Test_Helpers.cpp
 )
--- a/common/unit_test/CMakeLists.txt
+++ b/common/unit_test/CMakeLists.txt
@@ -19,11 +19,10 @@ add_executable(
 target_compile_features(unit-tests-kokkos-fft-common PUBLIC cxx_std_17)
 target_link_libraries(unit-tests-kokkos-fft-common PUBLIC common GTest::gtest)
 
-# Enable GoogleTest
-include(GoogleTest)
-gtest_discover_tests(unit-tests-kokkos-fft-common PROPERTIES DISCOVERY_TIMEOUT 600 DISCOVERY_MODE PRE_TEST)
-
 add_executable(unit-tests-kokkos-transpose Test_Main.cpp Test_Transpose.cpp)
 target_compile_features(unit-tests-kokkos-transpose PUBLIC cxx_std_17)
 target_link_libraries(unit-tests-kokkos-transpose PUBLIC KokkosFFT::fft GTest::gtest)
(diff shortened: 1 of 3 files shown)


common/unit_test/CMakeLists.txt
```diff
@@ -9,18 +9,21 @@ add_executable(
...
-
 target_link_libraries(unit-tests-kokkos-fft-common PUBLIC common GTest::gtest)
 
 # Enable GoogleTest
 include(GoogleTest)
 gtest_discover_tests(unit-tests-kokkos-fft-common PROPERTIES DISCOVERY_TIMEOUT 600 DISCOVERY_MODE PRE_TEST)
+
+add_executable(unit-tests-kokkos-transpose Test_Main.cpp Test_Transpose.cpp)
+target_compile_features(unit-tests-kokkos-transpose PUBLIC cxx_std_17)
+target_link_libraries(unit-tests-kokkos-transpose PUBLIC KokkosFFT::fft GTest::gtest)
```
science-enthusiast, line 24: These commands concerning the test executable could be moved before `include(GoogleTest)`. Then all the `gtest_discover_tests`s could be grouped together.

common/unit_test/Test_Transpose.cpp
```diff
@@ -49,1221 +50,294 @@ struct TestTranspose3D : public ::testing::Test {
...
   Kokkos::Random_XorShift64_Pool<execution_space> random_pool(12345);
   Kokkos::fill_random(exec, x, random_pool, 1.0);
   exec.fence();
 
-  for (int axis0 = 0; axis0 < DIM; axis0++) {
-    for (int axis1 = 0; axis1 < DIM; axis1++) {
-      for (int axis2 = 0; axis2 < DIM; axis2++) {
+  for (std::size_t axis0 = 0; axis0 < DIM; axis0++) {
+    for (std::size_t axis1 = 0; axis1 < DIM; axis1++) {
+      for (std::size_t axis2 = 0; axis2 < DIM; axis2++) {
```
science-enthusiast, line 322: Here and at other places, why this change to an unsigned int type?

yasahi-hpc: We have changed the definition of DIM from `int` to `std::size_t`.
Previously, it is a local `int` variable, but now it is a template variable in `std::size_t`

common/unit_test/CMakeLists.txt
yasahi-hpc: Good to know thanks
I will update this part

common/unit_test/Test_Transpose.cpp
science-enthusiast: ok. I see that now. Is it ok to make `DIM` an `int` type? In the tests, only integer literals like 1, 2, 3, ... are being passed as that parameter.

yasahi-hpc: Probably does not work with 
`using view_data_type = KokkosFFT::Impl::add_pointer_n_t<double, DIM>;`
which takes the `std::size_t` type as a template argument.

yasahi-hpc:
@science-enthusiast 
Thank you for your review
I have updated the CMake based on your review
Unfortunately, we need to define DIM in `std::size_t`

common/unit_test/CMakeLists.txt
science-enthusiast: Hi Yuuichi, I don't know whether multiple targets can be mentioned inside `gtest_discover_tests`. All I meant was to group all the `gtest_discover_tests` in one place for more readability.

yasahi-hpc: It works with multiple test cases
https://cmake.org/cmake/help/latest/module/GoogleTest.html#command:gtest_add_tests

common/unit_test/Test_Transpose.cpp
science-enthusiast: ok. How is `DIM` here related to `Rank` in the definition of `struct add_pointer_n`?

```
template <typename T, std::size_t Rank>
struct add_pointer_n { ... };
```

yasahi-hpc: I mean the type of the parameter `Rank` is `std::size_t`.
So, `DIM` needs to be `std::size_t` as well

science-enthusiast: Yes, I see that. I was wondering about the different names: DIM and Rank.

common/unit_test/CMakeLists.txt
science-enthusiast: ok

common/unit_test/Test_Transpose.cpp
yasahi-hpc: I often use `Rank` to `View` like data and `DIM` for others.
`add_pointer_n` is intended to create a data type of `View`

science-enthusiast: ok. Yes, I see how it recursively creates a type like `T*...*`.

yasahi-hpc:
@science-enthusiast 
Is this ready for merge?

science-enthusiast:
Can you point to the code that concerns this comment? "Also, a helper function to create a reference has also been refactored to avoid invalid usage of extent method." I see how `get_out_extents` helps in reducing duplicate logic.

common/unit_test/CMakeLists.txt
science-enthusiast: Hi Yuuichi, I read the documentation link. I am not sure where it is mentioned that multiple targets can be specified in `gtest_discover_tests`.

yasahi-hpc:
> Can you point to the code that concerns this comment? "Also, a helper function to create a reference has also been refactored to avoid invalid usage of extent method." I see how `get_out_extents` helps in reducing duplicate logic.

PR description rephrased. It is for `make_transposed` function which had out-of-rank View extent access previously.

common/unit_test/CMakeLists.txt
yasahi-hpc: I have fixed. For some reason, it worked in CI, but it is safer to have a single target

science-enthusiast (approved):
I looked at the new implementation of `make_transposed`.

Overall the PR looks good to me.

yasahi-hpc:
Thank you for the review
I will merge this after CI passes

PR #453, opened 2026-05-02
Move some util functions into Mapping and CheckConditions

yasahi-hpc:
This PR aims at moving some helper functions from `KokkosFFT_utils.hpp` into `KokkosFFT_Mapping.hpp` and `KokkosFFT_CheckConditions.hpp`. In the end, I would like to move all functions from `KokkosFFT_utils.hpp` to somewhere else and suppress `KokkosFFT_utils.hpp`

- [x] `are_aliasing`, `has_duplicate_values`, `is_out_of_range_value_included` and `are_valid_axes` are moved from `KokkosFFT_utils.hpp` into `KokkosFFT_CheckConditions.hpp`
- [x] `is_found`, `get_index` and `convert_negative_axes` are moved from `KokkosFFT_utils.hpp` into `KokkosFFT_Mapping.hpp`
- [x] Move corresponding tests as well
- [x] Update header includes

PR #455, opened 2026-05-04
Remove check in get_map_axes and remove axis API

yasahi-hpc:
Preparation for a bigger cleanup PR #453 

- [x] Remove `get_map_axes(const ViewType& view, int axis)` overload, which is unused
- [x] Remove unit-tests for this overload 
- [x] Remove `are_valid_axes` check in `get_map_axes` which should be made before calling this API


common/src/KokkosFFT_Mapping.hpp
daxmawal: > /// \throws if axes are not valid for the view

Maybe this is not in the PR scope.

/// \throws std::runtime_error if any axis is out of range
/// \pre axes must not contain duplicates after negative-axis conversion

yasahi-hpc: Thanks
I will update this part accordingly

daxmawal:
LGTM

yasahi-hpc:
Thank you for your review @daxmawal 
I will merge this

PR #462, opened 2026-05-12
Introduce pre-commit CI

yasahi-hpc:
This PR aims at replacing format check and the like with pre-commit.

- [x] Introduce `.pre-commit-config.yaml`
- [x] Move `clang-format-check`, `cmake-format-check`, and `spell-check` into `pre-commit`
- [x] Make `build_test` depends on `pre-commit-check`


.pre-commit-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - id: cmake-format
+
+    # Spell-checking
+  - repo: https://github.com/crate-ci/typos
+    rev: bbaefadf97b0ec5fdc942684b647f1a6ab250274 # v1.46.0
+    hooks:
+      - id: typos
+        name: Spell-check for typos
+        files: ./cmake/ ./CMakeLists.txt ./docs/ ./README.md ./common/ ./fft/ ./examples/ ./install_test/ ./testing/ ./distributed/
+        args: ['--config=./.typos.toml']
```
science-enthusiast, line 36: Are there directories which are not mentioned in this list?

Why I am asking is that for clang-format, there is no longer the need to mention directories and extensions.

yasahi-hpc: Maybe we can add .github directory which includes workflow yaml files
I think we need to specify directories for this typo check

.pre-commit-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - id: end-of-file-fixer
+      name: Fix EOFs
+      exclude: '^tpls/'
+    - id: check-yaml
+      name: Check YAML files
+    - id: check-added-large-files
+      name: Check adding large files
+
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: 579d9fe3b1ee34ddd6ebc5def531f2f1acce622e # v16.0.0
```
daxmawal: This changes the clang-format version from 17 to 16.0.0 compared to the previous workflow.

yasahi-hpc: That is a mistake
I will fix this

daxmawal (approved):

.pre-commit-config.yaml
yasahi-hpc: I only found v17.0.1, so I will use this

yasahi-hpc:
Thank you for your review and approval @daxmawal @science-enthusiast 
I will merge this

PR #469, opened 2026-05-20
Add nightly CI on ruche

yasahi-hpc:
This PR aims at adding a nightly test on ruche which is reported on [CDASH](https://my.cdash.org/index.php?project=Kokkos-FFT)

- [x] Add a general configuration file: CTestConfig.cmake
- [x] Add a nightly CI as a workflow

### DIFF
--- a/.github/workflows/nightly-ruche.yaml
+++ b/.github/workflows/nightly-ruche.yaml
@@ -5,6 +5,7 @@
 name: Nightly CEA builds
 
 on:
+  workflow_dispatch:
   schedule:
     - cron: "15 0 * * 1-5" # every weekday at midnight UTC
 
--- a/.github/workflows/nightly-ruche.yaml
+++ b/.github/workflows/nightly-ruche.yaml
@@ -6,7 +6,7 @@ name: Nightly CEA builds
 
 on:
   schedule:
-    - cron: "0 0 * * 1-5" # every weekday at midnight UTC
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-ruche.yaml
```diff
@@ -0,0 +1,131 @@
+# SPDX-FileCopyrightText: (C) The Kokkos-FFT development team, see COPYRIGHT.md file
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0 WITH LLVM-exception
+
+name: Nightly CEA builds
+
+on:
+  schedule:
```
daxmawal, line 9:
```suggestion
  workflow_dispatch:
  schedule:
```

Could we add [workflow_dispatch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatchrl) in addition to the scheduled trigger ?

This workflow is only triggered by [schedule](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule), it will only run once the workflow file is present on the default branch.

yasahi-hpc: Thanks
It would be useful

daxmawal (approved):

yasahi-hpc:
Thank you for approval @daxmawal 
I will merge this

PR #470, opened 2026-05-21
Add a const input API for hfft

yasahi-hpc:
As an experiment, this PR introduces an API to ensure that input data is unchanged. 

- [x] Update `hfft` function to manipulate const Views. If the input data is const Views, it makes an internal buffer and does not touch the input Views.
- [x] Add an unit-test that `hfft` does not update the const input Views

### DIFF
--- a/fft/src/KokkosFFT_Transform.hpp
+++ b/fft/src/KokkosFFT_Transform.hpp
@@ -216,13 +216,17 @@ void irfft(const ExecutionSpace& exec_space, const InViewType& in,
 /// \tparam InViewType: Input View type for the fft
 /// \tparam OutViewType: Output View type for the fft
 ///
-/// \param exec_space [in] Kokkos execution space
-/// \param in [in] Input data (real or complex)
-/// \param out [out] Output data (real)
-/// \param norm [in] How the normalization is applied (default, backward)
-/// \param axis [in] Axis over which FFT is performed (default, -1)
-/// \param n [in] Length of the transformed axis of the output (default,
+/// \param[in] exec_space Kokkos execution space
+/// \param[in,out] in Input data (real or complex)
+/// \param[out] out Output data (real)
+/// \param[in] norm How the normalization is applied (default, backward)
+/// \param[in] axis Axis over which FFT is performed (default, -1)
+/// \param[in] n Length of the transformed axis of the output (default,
 /// nullopt)
+/// \details The input data is modified in-place to save memory (if the input is
+/// non-const complex). If the input is real or const complex, we create a
+/// temporary complex view to perform the computation, and the input data is not
+/// modified.
 template <typename ExecutionSpace, typename InViewType, typename OutViewType>
 void hfft(const ExecutionSpace& exec_space, const InViewType& in,
           const OutViewType& out,
@@ -248,7 +252,9 @@ void hfft(const ExecutionSpace& exec_space, const InViewType& in,
 
   if constexpr (KokkosFFT::Impl::is_real_v<in_value_type> ||
                 std::is_const_v<typename InViewType::value_type>) {
-    // Convert to complex type if input is real
+    // For real input or const complex input, we create a temporary complex view
+    // to perform the computation, and the input data is not modified. Convert
+    // to complex type if input is real
     using ComplexType =
         std::conditional_t<KokkosFFT::Impl::is_real_v<in_value_type>,
                            Kokkos::complex<in_value_type>, in_value_type>;
@@ -266,7 +272,7 @@ void hfft(const ExecutionSpace& exec_space, const InViewType& in,
     Kokkos::deep_copy(exec_space, cin, in);
     if constexpr (!KokkosFFT::Impl::is_real_v<in_value_type>) {
       if (cin.span() >= std::size_t(std::numeric_limits<int>::max())) {
-        KokkosFFT::Impl::md_unary_operation<int64_t>(
+        KokkosFFT::Impl::md_unary_operation<std::int64_t>(
             "KokkosFFT::conjugate", exec_space, cin,
             KokkosFFT::Impl::Conjugate());
       } else {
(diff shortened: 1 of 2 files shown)


fft/src/KokkosFFT_Transform.hpp
```diff
@@ -246,9 +246,13 @@ void hfft(const ExecutionSpace& exec_space, const InViewType& in,
                      "axes are invalid for in/out views");
   auto new_norm = KokkosFFT::Impl::swap_direction(norm);
 
-  if constexpr (KokkosFFT::Impl::is_real_v<in_value_type>) {
-    // Convert to complex type
-    using ComplexType = Kokkos::complex<in_value_type>;
+  if constexpr (KokkosFFT::Impl::is_real_v<in_value_type> ||
+                std::is_const_v<typename InViewType::value_type>) {
+    // Convert to complex type if input is real
```
science-enthusiast: Similar to the comment below `else`, the comment here can be "For real input or const (real or complex) input, create a copy of complex type".

fft/src/KokkosFFT_Transform.hpp
```diff
@@ -258,21 +262,31 @@ void hfft(const ExecutionSpace& exec_space, const InViewType& in,
     auto extents     = KokkosFFT::Impl::extract_extents(in);
     ManageableComplexInViewType cin(
         "cin", KokkosFFT::Impl::create_layout<LayoutType>(extents));
+
     Kokkos::deep_copy(exec_space, cin, in);
+    if constexpr (!KokkosFFT::Impl::is_real_v<in_value_type>) {
+      if (cin.span() >= std::size_t(std::numeric_limits<int>::max())) {
```
science-enthusiast, line 274: We are implicitly assuming that `cin.span() < std::size_t(std::numeric_limits<std::size_t>::max())`. Is that condition checked already in some other place?

fft/src/KokkosFFT_Transform.hpp
```diff
@@ -258,21 +262,31 @@ void hfft(const ExecutionSpace& exec_space, const InViewType& in,
...
+        KokkosFFT::Impl::md_unary_operation<int64_t>(
```
science-enthusiast: `int64_t` can be `std::int64_t`, just like `std::size_t`.

fft/src/KokkosFFT_Transform.hpp
```diff
@@ -258,21 +262,31 @@ void hfft(const ExecutionSpace& exec_space, const InViewType& in,
...
+            "KokkosFFT::conjugate", exec_space, cin,
+            KokkosFFT::Impl::Conjugate());
+      } else {
+        KokkosFFT::Impl::md_unary_operation<int>("KokkosFFT::conjugate",
+                                                 exec_space, cin,
+                                                 KokkosFFT::Impl::Conjugate());
```
science-enthusiast, line 281: Instead of repeating the call to `KokkosFFT::Impl::md_unary_operation`, we could have a conditional type that will be either `std::int64_t` or `int` and pass that type to a single call to `KokkosFFT::Impl::md_unary_operation`.

fft/src/KokkosFFT_Transform.hpp
```diff
@@ -258,21 +262,31 @@ void hfft(const ExecutionSpace& exec_space, const InViewType& in,
...
       KokkosFFT::Impl::md_unary_operation<int64_t>(
-          "KokkosFFT::conjugate", exec_space, in_conj,
-          KokkosFFT::Impl::Conjugate());
+          "KokkosFFT::conjugate", exec_space, in, KokkosFFT::Impl::Conjugate());
     } else {
-      KokkosFFT::Impl::md_unary_operation<int>("KokkosFFT::conjugate",
-                                               exec_space, in_conj,
-                                               KokkosFFT::Impl::Conjugate());
+      KokkosFFT::Impl::md_unary_operation<int>(
+          "KokkosFFT::conjugate", exec_space, in, KokkosFFT::Impl::Conjugate());
```
science-enthusiast, line 293: Same as above, we can have a conditional type and single call to `KokkosFFT::Impl::md_unary_operation`.

fft/unit_test/Test_Transform.cpp
```diff
@@ -445,10 +447,32 @@ void test_fft1_1dhfft_1dview() {
...
+  // Check if the input is unchanged
+  KokkosFFT::hfft(exec, x_herm_const, out);
+  EXPECT_TRUE(allclose(exec, x_herm_const, x_herm_ref, 1.e-5, 1.e-6));
+  KokkosFFT::hfft(exec, x_herm_const, out_b,
+                  KokkosFFT::Normalization::backward);
+  EXPECT_TRUE(allclose(exec, x_herm_const, x_herm_ref, 1.e-5, 1.e-6));
+  KokkosFFT::hfft(exec, x_herm_const, out_o, KokkosFFT::Normalization::ortho);
+  EXPECT_TRUE(allclose(exec, x_herm_const, x_herm_ref, 1.e-5, 1.e-6));
+  KokkosFFT::hfft(exec, x_herm_const, out_f, KokkosFFT::Normalization::forward);
+  EXPECT_TRUE(allclose(exec, x_herm_const, x_herm_ref, 1.e-5, 1.e-6));
```
science-enthusiast: Since, we expect the const Views to be untouched, can we set a lesser value for the tolerance?

fft/src/KokkosFFT_Transform.hpp
yasahi-hpc: Sure

fft/src/KokkosFFT_Transform.hpp
yasahi-hpc: As the span of `Kokkos::View` is in `std::size_t`, we do not have that case.

fft/src/KokkosFFT_Transform.hpp
yasahi-hpc: OK

fft/src/KokkosFFT_Transform.hpp
yasahi-hpc: This conditional is a runtime check, so we need to instantiate both cases and choose one at runtime. 

`if (cin.span() >= std::size_t(std::numeric_limits<int>::max()))`

fft/unit_test/Test_Transform.cpp
yasahi-hpc: That is true. This should be exact

fft/src/KokkosFFT_Transform.hpp
science-enthusiast: Thanks for clarifiying. I should have thought about this being a run-time check.

yasahi-hpc:
Thanks @science-enthusiast for your review
I have updated based on your review
May I have another review

science-enthusiast (approved):
Looks good to me.

yasahi-hpc:
Thanks @science-enthusiast for approval
I will merge this after rebase

PR #480, opened 2026-05-27
Suppress warnings by explicitly define a return type

yasahi-hpc:
It seems that we get an error from the tuple construction. By explicitly specifying the type, warnings may be suppressed.
https://my.cdash.org/builds/3566534/build

### DIFF
--- a/common/src/KokkosFFT_Extents.hpp
+++ b/common/src/KokkosFFT_Extents.hpp
@@ -372,7 +372,12 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   }
   auto howmany = total_size(batch_extents);
 
-  return std::make_tuple(in_extents, out_extents, fft_extents, howmany);
+  using extents_type = std::vector<std::size_t>;
+  using return_type =
+      std::tuple<extents_type, extents_type, extents_type, decltype(howmany)>;
+
+  return return_type(std::move(in_extents), std::move(out_extents),
+                     std::move(fft_extents), howmany);
 }
 
 /// \brief Compute input, output and fft extents required for FFT
--- a/common/src/KokkosFFT_Extents.hpp
+++ b/common/src/KokkosFFT_Extents.hpp
@@ -5,10 +5,11 @@
 #ifndef KOKKOSFFT_EXTENTS_HPP
 #define KOKKOSFFT_EXTENTS_HPP
 
-#include <vector>
-#include <tuple>
 #include <algorithm>
 #include <numeric>
+#include <tuple>
+#include <utility>
+#include <vector>
 #include <Kokkos_Core.hpp>
 #include "KokkosFFT_Asserts.hpp"
 #include "KokkosFFT_CheckConditions.hpp"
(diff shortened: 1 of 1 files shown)


daxmawal:
LGTM

daxmawal (approved):

common/src/KokkosFFT_Extents.hpp
```diff
@@ -372,7 +373,12 @@ auto get_extents(const InViewType& in, const OutViewType& out,
   }
   auto howmany = total_size(batch_extents);
 
-  return std::make_tuple(in_extents, out_extents, fft_extents, howmany);
+  using extents_type = std::vector<std::size_t>;
+  using return_type =
+      std::tuple<extents_type, extents_type, extents_type, decltype(howmany)>;
+
+  return return_type(std::move(in_extents), std::move(out_extents),
+                     std::move(fft_extents), howmany);
```
science-enthusiast: Generally, `std::move` in a return statement is not needed and it can be counter-productive also. If you think the code will not look too ugly, the temporaries `slice_tail(in_extents_full)`, `slice_tail(out_extents_full)` and `slice_tail(fft_extents_full)` can be directly created here.

yasahi-hpc: Do you mean that `in_extents`, `out_extents`, and `fft_extents` in rvalue are considered as movable?

daxmawal: The current std::moves are not moving the returned object itself, only the vector elements into the tuple.

yasahi-hpc: Yes

science-enthusiast: Generally, one doesn't have to move a local object into the object that is being returned. Even though, C++ guarantees only RVO and not yet Named RVO, many compilers are capable of NRVO and a std::move in a return statement can break this optimization.

My sugestion for the return statement is this:
return return_type(slice_tail(in_extents_full), slice_tail(out_extents_full), slice_tail(fft_extents_full), howmany);

Since, slice_tail will be returning a temporary, these objects will be directly constructed inside the returned tuple.

daxmawal: I agree that constructing the tuple directly from the `slice_tail(...)` temporaries is cleaner here.

science-enthusiast: > The current std::moves are not moving the returned object itself, only the vector elements into the tuple.

I agree that the tuple that is getting created is a temporary. Thus, std::move of the arguments into its constructor should be fine. Still directly creating the constructor arguments as temporaries will be more efficient.

yasahi-hpc: OK, let's give it a try

yasahi-hpc:
Thanks @daxmawal @science-enthusiast 
I applied the suggestion from @science-enthusiast

common/src/KokkosFFT_Extents.hpp
science-enthusiast: > I agree that constructing the tuple directly from the `slice_tail(...)` temporaries is cleaner here.

I agree that saying it is a cleaner approach is more solid than saying it is a more efficient approach.

PR #491, opened 2026-06-24
[Docs] Add fftshift/ifftshift examples

yasahi-hpc:
This PR aims at adding a minimum working example in API reference. 

- [x] Add docs_fftshift.cpp which includes a minimum working example that is compiled in CI.
- [x] Add docs_ifftshift.cpp which includes a minimum working example that is compiled in CI.
- [x] Embed this in API reference with literal includes.


PaulGannay (approved):
LGTM

examples/docs/helper/docs_fftshift.cpp
```diff
@@ -0,0 +1,30 @@
...
+/// freqs: [0 1 2 3 4 5 -4 -3 -2 -1]
+/// KokkosFFT::fftshift(freqs);
+/// freqs_shifted: [-5 -4 -3 -2 -1 0 1 2 3 4]
+int main(int argc, char* argv[]) {
+  Kokkos::ScopeGuard guard(argc, argv);
+  using ExecutionSpace = Kokkos::DefaultExecutionSpace;
+
+  const int n0 = 10;
+  ExecutionSpace exec;
+  auto freq = KokkosFFT::fftfreq<ExecutionSpace, double>(exec, n0, 0.1);
```
Adrien-Tab: Can the function call be simplified ?
```suggestion
  auto freq = KokkosFFT::fftfreq(exec, n0, 0.1);
```

Adrien-Tab (approved):

examples/docs/helper/docs_ifftshift.cpp
```diff
@@ -0,0 +1,35 @@
...
+int main(int argc, char* argv[]) {
+  Kokkos::ScopeGuard guard(argc, argv);
+  using ExecutionSpace = Kokkos::DefaultExecutionSpace;
+  using ViewType       = Kokkos::View<double*, ExecutionSpace>;
+
+  ExecutionSpace exec;
+  ViewType freq_shifted("freq_shifted", 9);
+  auto h_freq_shifted = Kokkos::create_mirror_view(freq_shifted);
+  for (int i = 0; i < freq_shifted.extent_int(0); ++i) {
+    h_freq_shifted(i) = static_cast<double>(i - 4);
```
science-enthusiast, line 23: Is this static cast needed here? `int` should get implicitly converted to `double`.

docs/api/helper/fftshift.rst
```diff
@@ -7,3 +7,17 @@ KokkosFFT::fftshift
...
+.. literalinclude:: ../../../examples/docs/helper/docs_fftshift.cpp
+  :language: c++
+  :linenos:
+  :lines: 5-
+
+Expected output:
+
+.. code::
+
+ -5 -4 -3 -2 -1 0 1 2 3 4
```
science-enthusiast, line 23: Apart from specifying the expected output, could we describe what happens inside these example codes? What are the input frequencies to `KokkosFFT::fftshift` in this case and what are the inputs to `KokkosFFT::ifftshift` in the other case.

science-enthusiast (approved):
Apart from those minor suggestions, this looks good.

examples/docs/helper/docs_ifftshift.cpp
yasahi-hpc: Not necessarily
Just I often prefer explicit casting in this situation

science-enthusiast: ok. I prefer minimizing the use of static_casts and use them with comments wherever they are necessary (say, to overcome warnings about narrowing conversion).

docs/api/helper/fftshift.rst
yasahi-hpc: The behavior is explained in the docstrings saying

```
Shift the zero-frequency component to the center of the spectrum.
```

In my opinion, with this statement, we can understand what happens inside the example.
In the example, we say that

`[0 1 2 3 4 -5 -4 -3 -2 -1]` is shifted to `[-5 -4 -3 -2 -1 0 1 2 3 4]`

I fixed the mistake in the docstrings of the example

yasahi-hpc:
Thank you all for your review @Adrien-Tab @PaulGannay @science-enthusiast 
I will merge this after CI passes

docs/api/helper/fftshift.rst
science-enthusiast: Yes, the docstrings are sufficient.
Sorry, I still don't realize what the output of `KokkosFFT::fftfreq(10, 0.1);` will be like. But `0 1 2 3 4 -5 -4 -3 -2 -1` is more intuitively correct because there are equal number of positive and negative values, compared to `0 1 2 3 4 5 -4 -3 -2 -1`.

PR #517, opened 2026-08-20
Use concepts in Cuda plan wrapper

yasahi-hpc:
- [x] Use requires instead of SFINAE
- [x] Merge 1D-ND specializations of `create_plan` into ND one.

### DIFF
--- a/fft/src/KokkosFFT_Cuda_plans.hpp
+++ b/fft/src/KokkosFFT_Cuda_plans.hpp
@@ -5,6 +5,7 @@
 #ifndef KOKKOSFFT_CUDA_PLANS_HPP
 #define KOKKOSFFT_CUDA_PLANS_HPP
 
+#include <concepts>
 #include <Kokkos_Core.hpp>
 #include <Kokkos_Profiling_ScopedRegion.hpp>
 
@@ -18,9 +19,8 @@
 namespace KokkosFFT {
 namespace Impl {
 
-template <typename ExecutionSpace, typename T,
-          std::enable_if_t<std::is_same_v<ExecutionSpace, Kokkos::Cuda>,
-                           std::nullptr_t> = nullptr>
+template <typename ExecutionSpace, typename T>
+  requires(std::same_as<ExecutionSpace, Kokkos::Cuda>)
 void setup() {
   static bool once = [] {
     if (!(Kokkos::is_initialized() || Kokkos::is_finalized())) {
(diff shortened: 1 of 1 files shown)


fft/src/KokkosFFT_Cuda_plans.hpp
```diff
@@ -159,28 +67,40 @@ auto create_plan(const ExecutionSpace& exec_space,
...
 
   using index_type = FFTIndexType<Kokkos::Cuda>;
-  index_type idist = total_size(in_extents);
-  index_type odist = total_size(out_extents);
+  if constexpr (InViewType::rank() == 1 && fft_rank == 1) {
+    const index_type nx = fft_extents.at(0);
+    plan                = std::make_unique<PlanType>(nx, type, howmany);
+  } else if constexpr (InViewType::rank() == 2 && fft_rank == 2) {
+    const index_type nx = fft_extents.at(0), ny = fft_extents.at(1);
+    plan = std::make_unique<PlanType>(nx, ny, type);
```
science-enthusiast, line 75: Formatting is a little unexpected here.

fft/src/KokkosFFT_Cuda_plans.hpp
```diff
@@ -159,28 +67,40 @@ auto create_plan(const ExecutionSpace& exec_space,
       KokkosFFT::Impl::get_extents(in, out, axes, s, is_inplace);
 
   using index_type = FFTIndexType<Kokkos::Cuda>;
-  index_type idist = total_size(in_extents);
-  index_type odist = total_size(out_extents);
+  if constexpr (InViewType::rank() == 1 && fft_rank == 1) {
```
science-enthusiast, line 70: The previous interface was such that if `InViewType::rank()` is 1, 2 or 3, there was no need for `fft_rank` and `shape_type` was instantiated with 1, 2 or 3, respectively.

Earlier, it was implicitly assumed that `InViewType::rank() == fft_rank` if `InViewType::rank() <= 3`.

That being the case, these `if constexpr`s could just be based on `InViewType::rank()` being 1, 2 or 3.

yasahi-hpc: No, it was explicitly set as rank 1, 2 and 3. For example, 

```
// 3D transform
auto create_plan(const ExecutionSpace& exec_space,
                 std::unique_ptr<PlanType>& plan, const InViewType& in,
                 const OutViewType& out, Direction /*direction*/,
                 axis_type<3> axes, shape_type<3> s, bool is_inplace) {
```

science-enthusiast: Yes, I saw that. That is why I said there was no need for a variable called `fft_rank` when `InViewType::rank()` was 1, 2 or 3.

In that case, why check for both them being equal to 1, 2 or 3, now? Why not only check `InViewType::rank()`?

yasahi-hpc: Since we may have 2D FFT (rank 2) or 1D batched FFT (rank 1) on 2D Views.
`if constexpr (InViewType::rank() == 2 && fft_rank == 2)` only handles 2D FFT on 2D Views.

science-enthusiast: ok. I understand. Could you clarify to me, how "1D batched FFT (rank 1) on 2D Views" was handled before?

yasahi-hpc: We have overload for each before

1D batched FFT (rank 1) is handled by ND case and 2D FFT is handled by 2D case

science-enthusiast: Nevermind, I see that earlier the contraints were a combination of SFINAE based on `std::enable_if_t<InViewType::rank()` and function arguments with hardcoded template parameters.

science-enthusiast (approved):

yasahi-hpc:
Thank you for review and approval @science-enthusiast 
I will merge this

PR #532, opened 2026-09-04
Using bundled gtest if gtest or gmock is not found

yasahi-hpc:
On HPSF MI300A env, we could find `GTest::gtest` but cannot find `GTest::gmock`.
With this change, we use bundled gtest if gtest or gmock is not found 

[CI Failure log](https://my.cdash.org/builds/4039297/configure)


PaulGannay (approved):
LGTM

yasahi-hpc:
Thank you for approval @PaulGannay 
Let me merge this

